PR #4, opened 2016-01-29
Depend on "railties" instead of "rails".

amw:
Drops ActiveRecord and other dependencies for applications that only use parts of Rails suite.


awood45:
Thanks for this - I just finished testing that this, looks like we're safe to only use railties.

awood45:
There will be a delay on release though, because one of our tests depends on ActionMailer. I'm going to fix this on the test-only side, then push up.

awood45:
I've pushed this to Rubygems as version 1.0.1.

PR #6, opened 2016-04-19
Add secrets.json example

alifjamaluddin:
Add secrets.json example and tell user to add it in gitignore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,19 @@ for details. If you need to provide your own credentials, you can call
 client-creating actions manually. For example, to provide your own credentials
 for using Amazon Simple Email Service as a delivery method for ActionMailer:
 
+RAILS_ROOT/secrets.json
+```json
+{
+	"AccessKeyId":"your-access-key-id",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,9 +22,19 @@ for details. If you need to provide your own credentials, you can call
...
+	"SecretAccessKey":"your-secret-access-key"
+}
+```
+
+
+RAILS_ROOT/config/initializers/aws-sdk.rb
 ```ruby
 require 'json'
-# Because you're never going to commit credentials to source. Please.
+# Because you're never going to commit credentials to source. Please add it in .gitignore.
```
awood45: s/add it in/add it to

awood45:
I'm happy to take this, and the test failure is (unsurprisingly) unrelated. One grammatical change and I'll pull this in. Thanks!

README.md
anocriny: "Please add it into .gitignore"
This should be the right sentence is it?

awood45: Either works I think.

PR #8, opened 2016-05-24
Fix mailer example

stfnhrrs:
The former example looks to be from V1 of the AWS SDK, I've updated it to reflect changes in the V2 SDK


awood45:
The existing example is correct. Take a look at `aws-sdk-rails.rb` in the root lib directory to see what this is calling.

PR #10, opened 2016-06-11
to set message id to Mail::Message by ses response

genki0406:
action mailer deliver response is Mail::Message.
can't get ses mesage-id

### DIFF
--- a/lib/aws/rails/mailer.rb
+++ b/lib/aws/rails/mailer.rb
@@ -3,7 +3,7 @@ module Aws
 
     # Provides a delivery method for ActionMailer that uses Amazon Simple Email
     # Service.
-    # 
+    #
     # Once you have an SES delivery method you can configure Rails to
     # use this for ActionMailer in your environment configuration
     # (e.g. RAILS_ROOT/config/environments/production.rb)
@@ -31,7 +31,9 @@ module Aws
           send_opts[:destinations] = message.destinations
         end
 
-        @client.send_raw_email(send_opts)
+        result = @client.send_raw_email(send_opts)
+        message.message_id = "<#{result.message_id}@email.amazonses.com>"
+        result
 
       end
 
--- a/test/aws/rails/mailer_test.rb
+++ b/test/aws/rails/mailer_test.rb
@@ -26,7 +26,14 @@ module Aws
       def test_deliver
         message = sample_message
         resp = @mailer.deliver!(message)
-        assert_equal resp.context.params[:raw_message][:data].to_s, message.to_s
+
+        # over ride message-id by ses
+        message_id = resp.context.params[:raw_message][:data].to_s.match(/Message-ID: <(.*)>/)[1]
+        message_text = resp.context.params[:raw_message][:data].to_s.sub(
+          message_id,
+          "#{resp.message_id}@email.amazonses.com"
+        )
+        assert_equal message_text, message.to_s
         assert_equal resp.context.params[:destinations], message.destinations
       end
 


awood45:
I'm inclined to take this, or a variant of it. Can you tell me a bit more about the motivation of this and what you're wanting to use this for? Will help determine if this is the right change.

genki0406:
I think that's the right move is to return the actually transmitted content
It should reflect overwrite the contents if the SES is to overwrite

I want to use the message id for mail tracking
It troubled the correct message id can not get

genki0406:
how is it?
Please tell me if there is a point where not enough

Skudo:
I would love to have access to the SES message id, too.

When I send an email through `Aws::SES::Client#send_raw_email`, I can get the message id for that email by calling `#message_id` on the returned `Types::SendRawEmailResponse` object. The very same message id is used when I get SNS notifications for SES mails; you can find that message id in `notification['Message']['mail']['messageId']` and while it's possible to map SNS notifications to sent emails by other means, e. g. smart™ matching of other headers, using the message id for that purpose is the natural (and IMO right) way to do it.

In order to do that, we need to have access to the SES message id. I am not entirely certain if overwriting `message.message_id` is the right/best way, but that's for you to decide, @awood45. I'm just here to tell you why I want access to the SES message id. ;-D

awood45:
One problem with overwriting the `Mail::Message` `:message_id` trait is that it has [existing semantic meaning](http://www.rubydoc.info/github/mikel/mail/Mail/Message#message_id-instance_method), and that we don't necessarily want to just override it.

awood45:
I'm looking for alternative approaches right now.

Skudo:
I can't think of one off the top of my head, because the only thing we have right now is the `mail` object and modifying anything in there might cause some issues somewhere else. I'd have to go through `ActionMailer` code to see if there's something to jump in, but maybe you're about to tell me that won't help either.

Furthermore: Maybe I don't get it, because English is something like my 4.5th language, but what's the semantic meaning that makes us want to not override it? As far as I remember, backed by a bad memory, RFC 2822 only says that the `Message-ID` header value is a unique identifier for a given version of a given message. The `Mail::Message` implementation generates a message ID iff one isn't set already, which means that it's possible to send two messages with exactly the same message ID. Which in turns seems to be contradicting RFC 2822.

If we look at each mail sent handed over to SES as one new version of `message` each, I think setting `mail.message_id` to the SES message ID makes sense. Especially as the message ID generated by `Mail::MessageIdField#generate_message_id` isn't used when dispatching mails via SES, anyway.

Any mistake in my thought process (or somewhere else)?

ryanb:
I needed to track the `message_id` as well. I looked into `Mail` observers but unfortunately it does not send the response to the observers. My solution is to create a delivery method which allows a `tracker` option. This calls `delivered_email` just like a Mail Observer does except it passes the response as a second argument.

``` ruby
# config/initializers/ses_delivery_method.rb
class SesDeliveryMethod
  def initialize(options = {})
    @tracker = options[:tracker]
    @client = AWS::SES::Client.new(options.except(:tracker))
  end

  def deliver!(message)
    @client.send_raw_email(raw_message: {data: message.to_s}).tap do |response|
      @tracker.delivered_email(message, response) if @tracker
    end
  end
end

ActionMailer::Base.add_delivery_method(:ses, SesDeliveryMethod,
  region: ...,
  credentials: ...,
  tracker: SomeTracker.new
)

# config/environments/production.rb or any environment
config.action_mailer.delivery_method = :ses

# some_tracker.rb
class SomeTracker
  def delivered_email(message, response)
    # Do whatever you need to with response.message_id
  end
end
```

Creating your own delivery method is fairly simple. You can pretty easily customize it to your needs.

jaybloke:
I like your solution @ryanb. The only problem is that if you need to store the SES message_id with the record (for example, if you're tracking bounce, complaints etc), you're going to have to jump through some hoops to assign the message_id to the underlying AR model. Here's a slight tweak to your solution:

```
# config/initializers/ses_delivery_method.rb
class SesDeliveryMethod
  def initialize(options = {})
    @client = AWS::SES::Client.new(options)
  end

  def deliver!(message)
    @client.send_raw_email(raw_message: {data: message.to_s}).tap do |response|
      message.message_id = response.message_id
    end
  end
end

ActionMailer::Base.add_delivery_method(:ses, SesDeliveryMethod,
  region: ...,
  credentials: ...,
)

# config/environments/production.rb or any environment
config.action_mailer.delivery_method = :ses
```

This way, you can access the SES message_id directly from the mailer:

```
message = MyMailer.invite(model).deliver_now
model.message_id = message.message_id
model.save
```

PR #21, opened 2018-04-27
Update example to use Rails encrypted credentials

jonathanhefner:
Changed the example in the README to use Rails 5.2's encrypted credentials feature instead of a JSON file.

SIDE NOTE: I left the name of the delivery method as `:aws_sdk` for consistency with the example further down, and consistency with the default arg value of the `add_action_mailer_delivery_method` method, but I think the name `:aws_ses` would be more helpful.


awood45:
I like the new section, but I wouldn't want to remove the old section entirely since many customers are surely not on the latest version of Rails. If you can change this to have your new section be additive ("if you use Rails 5.2, you can use this feature instead"), happy to take this!

jonathanhefner:
@awood45 I brought back the original example, and added more details to both.  Let me know what you think.

PR #25, opened 2019-02-12
Record the SES Message ID in the header :ses_message_id

hughevans:
Following on from #10. Having the message ID is essential for tracking the deliverability (and other events).

**Description of changes:**

Instead of overriding `message_id` as was rejected in #10 I’ve placed the SES Message ID in a new header, so after you deliver a message through ActionMailer you have it available:

``` ruby
message = MyMailer.demo.deliver_now
message.header[:ses_message_id].value
>> "0000000000000000-1111111-2222-3333-4444-555555555555-666666"
```

This is adding the header after it’s sent only. I think it’s a reasonable conduit for passing this data back to the app in this instance.

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

### DIFF
--- a/test/aws/rails/mailer_test.rb
+++ b/test/aws/rails/mailer_test.rb
@@ -6,7 +6,17 @@ module Aws
     class MailerTest < Minitest::Test
 
       def setup
-        @mailer = Mailer.new(stub_responses: true)
+        @mailer = Mailer.new(
+          stub_responses: {
+            send_raw_email: {
+              message_id: message_id
+            }
+          }
+        )
+      end
+
+      def message_id
+        '0000000000000000-1111111-2222-3333-4444-555555555555-666666'
       end
 
       def sample_message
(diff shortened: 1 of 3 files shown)


awood45:
I actually like this plan. I'll timebox some time to brainstorm alternatives (say, tomorrow), and if I can't think of a better way, I'll merge and release this. I can't see a way that this would create problems, so I'm thinking I'm on board.

test/aws/rails/mailer_test.rb
```diff
@@ -25,9 +35,11 @@ def test_settings_method
 
       def test_deliver
         message = sample_message
-        resp = @mailer.deliver!(message)
-        assert_equal resp.context.params[:raw_message][:data].to_s, message.to_s
-        assert_equal resp.context.params[:destinations], message.destinations
+        data = @mailer.deliver!(message).context.params
+        body = data[:raw_message][:data].to_s
+        body.gsub!("\r\nHallo", "ses-message-id: #{message_id}\r\n\r\nHallo")
```
awood45: I'd actually like the test to mimic the behavior pattern you showed in the PR description. Perhaps retrieve the message body and message headers like a user would?

hughevans: Good idea, that’s pushed.

test/test_helper.rb
```diff
@@ -1,3 +1,18 @@
-require 'rails'
+require 'rails/railtie'
+require 'action_mailer'
 require 'aws-sdk-rails'
 require 'minitest/autorun'
+
+class TestMailer < ActionMailer::Base
+  layout nil
+
+  def deliverable(body:, from:, subject:, to:)
```
awood45: Please use an 'opts' hash rather than keyword args here. We still maintain test compat with older Ruby versions, for customers with LTS of deprecated versions, in this library. Those tests are breaking.

hughevans: No worries, that’s done.

awood45:
Thank you! I'll release this sometime tomorrow.

awood45:
Released as version `2.1.0`

PR #27, opened 2019-10-10
Drop support for end-of-life Ruby and Rails; Update CI

jonathanhefner:
This PR drops support for end-of-life versions of Ruby and Rails, and updates CI configuration to test against all supported versions.

This is a breaking change.  However, the previous CI configuration only tested against the latest release of Rails, so it is uncertain if the current version of this library works with older Rails versions, anyway.

I can amend this PR to include older versions of Rails, if required, but this seemed like a good opportunity to drop EOL versions.  If apps on older versions of Rails do work with the current major version of this library, they can still continue to use it.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .bundle
 .yardoc
 *.gem
+*.gemfile.lock
 coverage
 doc
 Gemfile.lock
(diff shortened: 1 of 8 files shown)


mullermp:
I would like to drop support for old rubies, too. I had started work in a 'v3' branch. I would like to merge these, however, I am thinking of having them sit in master without a changelog so that our build system doesn't do a minor version bump. I would like to get a bunch of changes in (including this) for a new major version.

PR #29, opened 2019-10-16
V3 preparation

mullermp:
Tracking changes for V3 (so far):

Adds rubocop / fix violations
Drops support for Ruby <= 2.3
Drops support for Rails <= 5.1
Adds credential secrets
Dummy test app
Changes :aws_sdk to :ses in preparation for potential pinpoint support


mullermp:
Some extra work needs to be done, including tests for credentials.

mullermp:
@jonathanhefner for visibility

.travis.yml
```diff
@@ -4,14 +4,16 @@ rvm:
   - 2.4
   - 2.5
   - 2.6
+  - 2.7
```
jonathanhefner: You may want to add an `allow_failures` sections for Ruby 2.7, since it is still in development.

.travis.yml
```diff
@@ -4,14 +4,16 @@ rvm:
...
+  - ruby
```
jonathanhefner: Does this just test with the latest official release of Ruby?  If so, you may want to add an `allow_failures` section for this too.  Although it's good to know if a new Ruby release breaks old code, you probably don't want that to fail the builds of incoming PRs.

Or possibly just remove this, and manually add new Ruby versions.  You probably want to explicitly list all supported versions to prevent gaps in coverage, anyway.

CHANGELOG.md
```diff
@@ -4,17 +4,27 @@ Unreleased Changes
...
-* Issue - Ensure `aws-sdk-rails.initialize` executes before `load_config_initializers`
+* Issue - Ensure `aws-sdk-rails.initialize` executes before
+`load_config_initializers`
 
 2.0.0 (2017-08-29)
 ------------------
 
-* Upgrading - Support version 3 of the AWS SDK for Ruby. This is being released as major version 2 of `aws-sdk-rails`, though the APIs remain the same. Do note, however, that we've changed our SDK dependency to only depend on `aws-sdk-ses`. This means that if you were depending on other service clients transitively via `aws-sdk-rails`, you will need to add dependencies on the appropriate service gems when upgrading. Logger integration will work for other service gems you depend on, since it is wired up against `aws-sdk-core` which is included in the `aws-sdk-ses` dependency.
+* Upgrading - Support version 3 of the AWS SDK for Ruby. This is being released
+as major version 2 of `aws-sdk-rails`, though the APIs remain the same. Do note,
```
jonathanhefner: Nitpick: setting the wrap margin to 72 characters (instead of 80) can make side-by-side diffs easier to read on screens with less real estate.

Rakefile
```diff
@@ -1,16 +1,22 @@
-require "bundler/gem_tasks"
+# frozen_string_literal: true
+
+require 'bundler/gem_tasks'
```
jonathanhefner: The `aws-sdk-ruby` repo appears to prefer single quotes, so I guess it makes sense to switch.  But, just FYI, Rails itself and many projects in the Rails ecosystem prefer double quotes ([reference](https://github.com/rubocop-hq/rubocop/issues/5306#issuecomment-354449447)).

aws-sdk-rails.gemspec
```diff
@@ -1,20 +1,22 @@
...
-  spec.authors       = ["Amazon Web Services"]
-  spec.email         = ["alexwood@amazon.com"]
-  spec.summary       = "AWS SDK for Ruby Rails Plugin"
-  spec.description   = "Provides helpers to integrate the AWS SDK for Ruby with Ruby on Rails."
-  spec.homepage      = "https://github.com/aws/aws-sdk-rails"
-  spec.license       = "Apache 2.0"
+  spec.authors       = ['Amazon Web Services']
+  spec.email         = ['chejingy@amazon.com', 'mamuller@amazon.com']
+  spec.summary       = 'AWS SDK for Ruby on Rails Plugin'
+  spec.description   = 'Integrates the AWS Ruby SDK with Ruby on Rails'
```
jonathanhefner: Love it! :heart:

lib/aws/rails/mailer.rb
```diff
@@ -1,21 +1,21 @@
...
+    #
     # Once you have an SES delivery method you can configure Rails to
     # use this for ActionMailer in your environment configuration
     # (e.g. RAILS_ROOT/config/environments/production.rb)
     #
-    #     config.action_mailer.delivery_method = :aws_sdk
+    #     config.action_mailer.delivery_method = :ses
     #
-    # Uses the AWS SDK for Ruby V2's credential provider chain when creating an
+    # Uses the AWS SDK for Ruby V3's credential provider chain when creating an
```
jonathanhefner: Maybe just remove the "V3's" altogether, so that it can't fall out of date in the future?

lib/aws_sdk_rails.rb
```diff
@@ -1,34 +1,31 @@
...
     # @api private
     class Railtie < ::Rails::Railtie
-      initializer "aws-sdk-rails.initialize", before: :load_config_initializers do |app|
+      initializer 'aws_sdk_rails.initialize',
+                  before: :load_config_initializers do
         # Initialization Actions
-        Aws::Rails.use_rails_encrypted_credentials
         Aws::Rails.add_action_mailer_delivery_method
         Aws::Rails.log_to_rails_logger
+        Aws::Rails.use_rails_encrypted_credentials
```
jonathanhefner: Does this need to go before `add_action_mailer_delivery_method`?  Or does the mailer use the newly-added credentials regardless of order?  Should there be another test to verify this?

jonathanhefner:
@mullermp I left a few minor comments, but this looks great!  I'm excited to see this move forward.  I will take a closer look at the credentials configuration issue tomorrow.  Cheers! :tada:

mullermp:
@jonathanhefner Thanks for all of the comments!! I will work through those.

.travis.yml
mullermp: Good call. I think I'll just remove it for now and I can add it later. What do you think about also removing `jruby`? Historically it's been a nightmare in `aws-sdk-ruby` with breaking tests. I wouldn't say that it's officially supported by us at this point.

lib/aws/rails/mailer.rb
mullermp: Good call. I am going to use language such as 'latest version'.

lib/aws_sdk_rails.rb
mullermp: I get in the habit of alphabetizing methods and method calls - but that is a good point. This *should* be first, otherwise, the client can be constructed using the provider chain instead of encrypted credentials. I will fix this.

jonathanhefner:
Regarding the credentials configuration issue, what's happening is that the paths for the credential files are initialized before we override the Rails root in `test_helper.rb`.  We can override these file paths too, but this kind of issue may bite us again in the future, so here is (I think) a better solution:

1. Move the entire `module Dummy` from `test/test_helper.rb` to `test/dummy/config/application.rb`, and remove the `config.root = ...` line.
2. Put `require_relative "dummy/config/application.rb"` in `test/test_helper.rb` where `module Dummy` was, below `ENV["RAILS_ENV"] = "test"`.
3. Create a file `test/dummy/config.ru` with following contents: `# This file indicates the Rails root directory`

.travis.yml
jonathanhefner: In theory, JRuby is great...

It's not officially supported by Rails either though.  Perhaps you can add an `allow_failures` section for it.  It looks like `aws-sdk-ruby` does this, and I've read it's helpful to people who do use JRuby, so they at least know the status of your library.

Rakefile
mullermp: Thanks for the reference! That was an interesting read and some good insight into what the community prefers. I prefer double quotes myself but I'm more keen on taking whatever the defaults are (which is currently single quotes, as well as 80 line characters). Our team manages many repos (with the bigger one being aws-sdk-ruby) so having defaults makes my life a bit easier in some fashion.

.travis.yml
mullermp: Ah interesting. If rails doesn't officially support it - I think it's okay to make it unofficially supported here too; after all, this is supposed to be a rails plugin package. Though I would expect JRuby users wouldn't run into any issues using this anyway. For now, I'll keep jruby and add an allowed failure if it becomes a problem.

mullermp:
You mentioned cloudwatch support. What do you think about adding code similar to [this package](https://github.com/zshannon/cloudwatchlogger)? Of course I will reach out to the current maintainer and check licensing.

mullermp:
Touched up the README as well as other things. Please take another look at your convenience!

I am thinking other features can be added as minor versions and that this is looking good for a V3 as is, pending some minor organizational changes.

test/test_helper.rb
```diff
@@ -6,13 +6,6 @@
...
 
-module Dummy
-  class Application < Rails::Application
-    config.root = File.join(__dir__, 'dummy')
-    config.load_defaults Rails::VERSION::STRING.to_f
-    config.eager_load = false
-    config.require_master_key = true
-  end
-end
+require_relative 'dummy/config/application.rb'
```
jonathanhefner: D'oh! I copied and pasted the path in my comment, and I forgot to remove the `.rb`.  It's not necessary here.

README.md
```diff
@@ -70,9 +90,8 @@ Aws::Rails.add_action_mailer_delivery_method(
 )
 ```
 
-Or, if you are storing your AWS keys using Rails 5.2's [Encrypted
-Credentials](http://guides.rubyonrails.org/security.html#custom-credentials),
-use the following initializer code instead:
+Or, if you are using Rails 5.2's Encrypted Credentials, use the following
```
jonathanhefner: I think this section can be dropped now since encrypted credentials are loaded by default, and the example above this one does a good job of showing how to override.

mullermp: I can do that - perhaps I can condense those down to one as appropriate.

jonathanhefner:
Everything looks good! :+1: 

> I am thinking other features can be added as minor versions and that this is looking good for a V3 as is, pending some minor organizational changes.

I agree.

With regards to CloudWatch integration, something similar to that gem would be good.  Plus maybe a sleek way to initialize.  For example, instead of having to set:
```ruby
Rails.application.config.logger = Aws::Rails::CloudWatchLogger.new(
  # ...
)
```
The railtie could check for something like `Aws.config[:default_cloudwatch_log_group]`, and automatically set up the logger if it finds it.

mullermp:
Excellent. Thanks for the code reviews. I will major version bump this when I can.

CHANGELOG.md
```diff
@@ -4,17 +4,27 @@ Unreleased Changes
 2.1.0 (2019-02-14)
```
cjyclaire: Make sure also prepare for the changelog entry, you can make it a list of`* Feature -`, versions and date title will be auto bumped later based on `VERSION` file.

Thus, also need to include a update for VERSION file as well : D

test/test_helper.rb
```diff
@@ -1,11 +1,7 @@
 # frozen_string_literal: true
 
-require 'rails/all'
-require 'aws_sdk_rails'
 require 'minitest/autorun'
 
-ENV['RAILS_ENV'] = 'test'
```
jonathanhefner: Note that although this line is not required for our current purposes, it will be required if any `ActionDispatch::IntegrationTest` test classes are added in the future.  (See jonathanhefner/garden_variety@55af1a2 for an explanation.)

mullermp: Ah ok; added back, then.

CHANGELOG.md
mullermp: Done - under the guise of "Upgrading" similar to the 2.0.0 bump.

mullermp:
@jonathanhefner Just FYI - this has been released as v3.0.3. Don't ask what happened to v3.0.0/2 :)

Issue #32, opened 2020-04-03
Rails.application.credentials.aws reserved for aws credentials 👎


majkelcc:
Recently in our application that uses this gem, we migrated from previous `Rails.secrets` mechanism to the new `Rails.application.credentials`. Unfortunately, after moving the same keys from `config/secrets.yml.enc` to new encrypted credentials we saw an error about `invalid configuration option: ':rds_instance'`, which we stored under `Rails.application.credentials.aws`.

I tried to look in `aws-sdk-ruby` for code that automatically loads these credentials. With no luck there, I assumed it was Rails itself that's responsible. I was very happy to see that [this was not the case](https://github.com/rails/rails/issues/38754#issuecomment-602276489).

Reserving keys in the standard credentials mechanism is bad for many reasons::
1. It's magic and unexpected
2. Was not present in previous `Rails.application.secrets` making the migration difficult
3. This `invalid configuration option` error is difficult to debug (validation happens late in a different gem and `aws-sdk-rails` does not appear in the stack trace)
4. Most importantly, if more gems would start reserving their own keys in `Rails.application.credentials`, it would be a mess

I would want to call for this behavior to be deprecated and eventually removed from defaults.

alextwoods:
I've been looking into this and agree that the current process of loading all values under the `aws` namespace in the Rails Encrypted Credentials is a bad practice.  I've created a PR to address this: #33.  This change prevents the "aws" namespace from being "reserved" while still allowing users of this gem to have their aws credentials loaded.

PR #33, opened 2020-04-03
Only merge aws credentials from Rails Encrypted Credentials.

alextwoods:
*Issue:*
Fixes: #32 

*Description of changes:*
Only merge aws credential related keys from the aws namespace in Rails Encrypted Credentials.


majkelcc:
@alextwoods, thanks for prompt action. Although better, I think that automatically loading these credentials does more harm than good. This is a niche gem doing simple things and for it to affect something as important as loading AWS configuration sounds like trouble. I'll be dropping my dependency on this gem because I'm afraid of ideas like this 😬

Since this is a breaking change already, please consider making this whole feature an opt-in. Thank you and stay safe 😇🖖

alextwoods:
Thanks for the feedback - loading of AWS credentials is one of the main features of this Gem and something that other users had requested though I appreciate the concern around the usage of the namespace.

Issue #14, opened 2017-03-13
Feature Request: ActiveSupport Notifications


guyisra:
Hi,

In the current aws-sdk-rails and in aws-sdk-ruby there is no instrumentation support (unless I've missed it?) 

Is it possible to publish an ActiveSupport::Notification for every AWS call so we can subscribe to these events?

thanks

awood45:
Can definitely take this as a feature request as we start investing in more features. Instrumentation support is something on our roadmap for this year, and this is one possible route.

sivagollapalli:
Hi, I can take a look into this feature request if anyone could provide guidance about where to start and end goal?

mullermp:
Hi @sivagollapalli, thanks for your interest. It looks like the original request was to publish an ActiveSupport notification whenever the SDK is invoked. I'm not sure how you would go about that without taking a look myself. Another use case (not mentioned) might be to have an ActiveSupport::SNS notification of some sort.

As a user of `aws-sdk-rails` what kinds of features would you like to see?

sivagollapalli:
@mullermp Frankly speaking I don't know the use cases of this feature 😄 . In my opinion lets add instrumentation for one or two popular services like s3 or Ec2 and check how developers are using this and then establish to remaining services. let me know your thoughts.

mullermp:
ActiveSupport has a guide here: https://edgeguides.rubyonrails.org/active_support_instrumentation.html#creating-custom-events

I think that we can instrument every client call, perhaps with "<service-name>.<operation>"  (ex: "s3.list_buckets")

My thought is that we should be agnostic to any client for this to be feature complete. The part that is missing (and requires investigation) would be how to detect calls to any of the Aws clients.

sivagollapalli:
@mullermp Ok. let me do some investigation.

PR #37, opened 2020-11-07
ActiveSupport::Notifications instrumentation of aws sdk client calls

alextwoods:
Fixes: #14 

*Description of changes:*
Support ActiveSupport::Notification instrumentation of all AWS SDK client operations.  

This is currently not enabled by default and requires calling `Aws::Rails.instrument_sdk_operations` (eg, can be done in config/application.rb or in environment configs).  

Instrumentation adds ~0.2ms per call (which is an ~20% increase outside of network calls).

Note: I had to inspect all loaded Aws modules for service modules and add the handler to each one - adding to the seahorse base client does not work at this stage.

Open Questions:
1. Should this be enabled/disabled by config? And if so, should we allow configuration of which clients are instrumented.  Answer: Removed config.  Instrumentation is added by calling `Aws::Rails.instrument_sdk_operations` (See above explanation)
2. What should the name of the event be?  Answer: `<operation>.<serviceId>.aws` (Note, this naming matches the [patterns used in rails](https://edgeguides.rubyonrails.org/active_support_instrumentation.html)).
3. Is there a better way to do this than adding a plugin/handler? Answer: No, this is the best approach here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,19 +120,19 @@ message['X-SES-FROM-ARN'] = 'arn:aws:ses:us-west-2:012345678910:identity/bigchun
 message.deliver
 ```
 
-## Adding `ActiveSupport::Notifications` Instrumentation to AWS SDK calls
+## Active Support Notification Instrumentation for AWS SDK calls
 To add `ActiveSupport::Notifications` Instrumentation to all AWS SDK client 
 operations call `Aws::Rails.instrument_sdk_operations` before you construct any
 SDK clients.
 
-Example usage in `config/capplication.rb`
+Example usage in `config/initializers/instrument_aws_sdk.rb`
 ```ruby
 Aws::Rails.instrument_sdk_operations
 ```
 
 Events are published for each client operation call with the following event
 name: <operation>.<serviceId>.aws.  For example, S3's put_object has an event
-name of: put_object.S3.aws.  The payload of the event is the 
+name of: `put_object.S3.aws`.  The payload of the event is the 
 [request context](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Seahorse/Client/RequestContext.html).
 
 You can subscribe to these events as you would other
--- a/README.md
+++ b/README.md
@@ -119,3 +119,32 @@ message = MyMailer.send_email(options)
 message['X-SES-FROM-ARN'] = 'arn:aws:ses:us-west-2:012345678910:identity/bigchungus@memes.com'
 message.deliver
 ```
+
+## Adding `ActiveSupport::Notifications` Instrumentation to AWS SDK calls
+To add `ActiveSupport::Notifications` Instrumentation to all AWS SDK client 
+operations call `Aws::Rails.instrument_sdk_operations` before you construct any
+SDK clients.
+
+Example usage in `config/capplication.rb`
+```ruby
+Aws::Rails.instrument_sdk_operations
+```
+
+Events are published for each client operation call with the following event
+name: <operation>.<serviceId>.aws.  For example, S3's put_object has an event
+name of: put_object.S3.aws.  The payload of the event is the 
+[request context](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Seahorse/Client/RequestContext.html).
+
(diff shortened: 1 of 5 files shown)


lib/aws-sdk-rails.rb
```diff
@@ -47,5 +49,19 @@ def self.use_rails_encrypted_credentials
           .to_h.slice(*aws_credential_keys)
       )
     end
+    
+    def self.add_notifications_instrument_plugin
+      # TODO: Config based.  Enable/disable.
+      # specify which clients get instrumented?
```
mullermp, line 55: I'd say config on or off and nothing else

lib/aws/rails/notifications_instrument_plugin.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      def add_handlers(handlers, config)
+        handlers.add(Handler, step: :initialize)
+      end
+
+      class Handler < Seahorse::Client::Handler
+
+        def call(context)
+          # TODO: Should this be #{operation_name}.#{service}?
+          ActiveSupport::Notifications.instrument('operation.aws_sdk', context: context) do
```
mullermp, line 18: "aws.#{client.class}.#{operation_name}"

because notification subscriptions can use regex

lib/aws/rails/notifications_instrument_plugin.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'aws-sdk-core'
+
+module Aws
+  module Rails
+    # @api private
+    class NotificationsInstrument < Seahorse::Client::Plugin
```
mullermp, line 8: I'm not a fan of the name.. how about just Notifications (already under Aws::Rails)

README.md
```diff
@@ -119,3 +119,32 @@ message = MyMailer.send_email(options)
 message['X-SES-FROM-ARN'] = 'arn:aws:ses:us-west-2:012345678910:identity/bigchungus@memes.com'
 message.deliver
 ```
+
+## Adding `ActiveSupport::Notifications` Instrumentation to AWS SDK calls
```
mullermp, line 123: nit, title this "Active Support Notification Instrumentation for AWS SDK calls"

README.md
```diff
@@ -119,3 +119,32 @@ message = MyMailer.send_email(options)
...
+To add `ActiveSupport::Notifications` Instrumentation to all AWS SDK client 
+operations call `Aws::Rails.instrument_sdk_operations` before you construct any
+SDK clients.
+
+Example usage in `config/capplication.rb`
+```ruby
+Aws::Rails.instrument_sdk_operations
+```
```
mullermp, line 131: let's prefer to make an initializer here

lib/aws/rails/notifications_instrument_plugin.rb
alextwoods: Yeah good point here.  I went with <operation>.<service>.aws (following the rails convention from https://edgeguides.rubyonrails.org/active_support_instrumentation.html).

lib/aws/rails/notifications_instrument_plugin.rb
alextwoods: Wanted to differentiate between pub and sub here, but agree the name is long and awkward.

README.md
```diff
@@ -119,3 +119,32 @@ message = MyMailer.send_email(options)
...
+
+Events are published for each client operation call with the following event
+name: <operation>.<serviceId>.aws.  For example, S3's put_object has an event
+name of: put_object.S3.aws.  The payload of the event is the
```
mullermp, line 135: wrap in ``

lib/aws-sdk-rails.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 require_relative 'aws/rails/mailer'
+require_relative 'aws/rails/notifications_instrument_plugin'
```
mullermp, line 4: you're going to need to rebase this against my change once merged, or vice versa, up to you.

mullermp (approved):

PR #38, opened 2020-11-10
Add DynamoDB session store (v2 session store gem)

mullermp:
Uses a future version of aws-sessionstore-dynamodb that removed Rails support and adds it here.

Includes a generator for a migration `rails generate dynamo_db:session_store_migration`

Includes a rake task for cleanup (outside TTL) `rake dynamo_db:session_store:clean`

Supports loading environment based configuration `dynamo_db_session_store.yml`

Ref https://github.com/aws/aws-sessionstore-dynamodb-ruby/pull/26

### DIFF
--- a/lib/aws/rails/railtie.rb
+++ b/lib/aws/rails/railtie.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module Aws
+  # Use the Rails namespace.
+  module Rails
+    # @api private
+    class Railtie < ::Rails::Railtie
+      initializer 'aws-sdk-rails.initialize',
+                  before: :load_config_initializers do
+        # Initialization Actions
+        Aws::Rails.use_rails_encrypted_credentials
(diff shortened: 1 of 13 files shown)


lib/aws/rails/railtie.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Aws
+  # Use the Rails namespace.
+  module Rails
+    # @api private
+    class Railtie < ::Rails::Railtie
+      initializer 'aws-sdk-rails.initialize',
```
alextwoods, line 8: Not related to this change, but agree we should move each to a seperate initializer.

alextwoods (approved):

lib/aws/rails/railtie.rb
mullermp: Agreed, but I need to know which initializer should go when.

alextwoods (approved):

PR #40, opened 2020-11-12
Sessions testing

mullermp:
Adds testing for dynamodb session store


test/dummy/config/dynamo_db_session_store.yml
```diff
@@ -0,0 +1,70 @@
+# Uncomment and manipulate the key value pairs below
```
alextwoods, line 1: Should this not be checked in?  Assume its getting generated by the tests

test/dummy/db/migrate/20201112193021_create_dynamo_db_sessions_table.rb
```diff
@@ -0,0 +1,9 @@
+class CreateDynamoDbSessionsTable < ActiveRecord::Migration[6.0]
```
alextwoods, line 1: Ditto on this - I think we'd end up with a new file each time from the test w/ differnt date stamps?

alextwoods (approved):

test/dummy/db/migrate/20201112193021_create_dynamo_db_sessions_table.rb
mullermp: Rails is smart enough to check the contents (and sometimes the name) so even running at a later date will determine that the file is identical and not create a new one.

test/dummy/config/dynamo_db_session_store.yml
mullermp: I dumped these files in /dummy/ so they could be used with the config loading tests (action_dispatch/session)

alextwoods (approved):

PR #42, opened 2020-11-17
SQS Active job

alextwoods:
*Description of changes:*
A lightweight, SQS backend for ActiveJob.

There are three components:
1.  An SQS Active Job Adapter.  Responsible for serializing the job and sending it to the right queue.
2. Configuration.  Contains mapping of Active Job queue names to SQS Queue URls (and other settings) - shared between 1 and 3.
3. A Poller/Executor  - Polls for messages on the queue, unmarshals jobs and runs them (using a loaded rails environment).  

TODOs:
* ~~Library/feature naming~~:  A: AWS SQS Active Job (adapter name will be `:aws_sqs`).
* ~~Poller Logging support~~  A:   Uses Rails.logger by default, but logger can be configured.
* ~~Which signals should be trapped and handled as graceful shutdown?~~" A: `SIGINT` and `TERM` are trapped for clean shutdown.
* ~~Tests~~
* ~~optimize default polling arguments (and add additional config parameters)~~


mullermp (approved):
Nice work!

PR #43, opened 2020-11-24
Initial task and generators (from aws-record-generator-rails)

mullermp:


lib/generators/aws_record/model/templates/model.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+<% end -%>
+<% if options['table_name'] -%>
+  set_table_name "<%= options['table_name'] %>"
+<% end -%>
+<% if options.key? :password_digest -%>
+  has_secure_password
+<% end -%>
+end
+<% end -%>
```
alextwoods, line 74: Add a new line.

lib/aws-sdk-rails.rb
```diff
@@ -5,3 +5,7 @@
 require_relative 'aws/rails/notifications'
 
 require_relative 'action_dispatch/session/dynamodb_store'
+
+require_relative 'generators/aws_record/base'
+require_relative 'generators/aws_record/generated_attribute'
```
alextwoods, line 10: You should be able to skip these here (since they are required in the base).

lib/generators/aws_record/base.rb
```diff
@@ -0,0 +1,218 @@
...
+        private
+
+        def initialize(args, *options)
+          options[0] << "--skip-table-config" if options[1][:behavior] == :revoke
+          @parse_errors = []
+
+          super
+          ensure_unique_fields
+          ensure_hkey
+          parse_gsis!
```
alextwoods, line 33: I'm not sure why some of these methods are `!` and others are not.  I'd argue for consistency, just leave them out and make all of them private.

alextwoods (commented):
Looks good overall.  The migrated base could use a lot of love, but outside the scope of this pr.

alextwoods (approved):

Issue #54, opened 2021-01-07
Support for forwarding Elastic Beanstalk SQS Daemon requests to Active Job


MrWellington:
Hi everyone, just discovered this gem recently. So far has been a great productivity tool especially as someone new to Rails and AWS, appreciate the good work!

I am currently hosting an application via Elastic Beanstalk and using your Active Job queue adapter to send messages to SQS. To process the queue I'm looking to host the app in an EB worker environment. Those instances come with an SQS Daemon that retrieves messages and forwards them to localhost over HTTP. Instead of using the provided **_aws_sqs_active_job_** process to retrieve messages (which I believe would conflict with the daemon), I would like the ability to listen for messages from the daemon and forward them to the appropriate jobs.

The existing library [Active Elastic Job](https://github.com/tawan/active-elastic-job) provides this capability using a lightweight Rack middleware that is only added with the presence of a specific environment variable. However that gem is largely out of date (still requires version 2.x of the core SDK) and requires the use of their queuing backend as well (I'd rather use yours). 

I thought the approach of using middleware to achieve this was sound, and for the time being I currently have a stripped-down version of that library running successfully in my source code. However I thought this could be a nice addition to the Rails SDK itself that others could benefit from. Is this an idea you would be willing to implement or accept contributions for?

alextwoods:
Great to hear its been useful!  We already support using AWS Lambda to process ActiveJobs and I believe this feature-request is very thematically similar and something we should support.  I agree using rack middleware seems like a sound implementation, and should be fairly straightforward.

We'd be very happy to accept contributions for this.  If you plan on implementing, let me know, otherwise, I'll add this to my backlog and implement in the next couple of weeks.

MrWellington:
Glad to hear it, and thanks for the quick response!

Should have time to work on an implementation over the next few days, will try to send out a request soon and keep you posted one way or the other.

PR #55, opened 2021-01-14
Support for processing Elastic Beanstalk worker SQS Daemon requests in Active Job

MrWellington:
For [feature-request 54](https://github.com/aws/aws-sdk-rails/issues/54):

Here's a first pass at an implementation for this feature. In short the intent is for the middleware to be added in the presence of an "AWS_PROCESS_BEANSTALK_WORKER_REQUESTS" environment variable (can change if that violates any conventions). This middleware should be able to retrieve Active Job SQS messages forwarded to the application by the daemon and pass them to their associated jobs. There is also code to handle scheduled jobs created by the worker environment itself. More information can be found in the readme updates.

Note that the [Active Elastic Job](https://github.com/tawan/active-elastic-job) gem that influenced this design involved an additional security measure wherein the queuing backend set a message digest -- and the middleware verified it -- using the same Rails SECRET_KEY_BASE value in both environments. I didn't consider this to be essential (thought restricting senders to localhost/docker host was more than adequate), and to reduce impact to the existing code did not implement it. Wanted to get it on your radar though - if you agree it is not essential at present might be a nice additional security layer to add later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,9 +362,10 @@ Another option for processing jobs without managing the worker process is hostin
 This SDK includes Rack middleware that can be added conditionally and which will process requests from the 
 SQS Daemon provided with each worker instance. The middleware will forward each request and parameters to their appropriate jobs.
 
-To protect against forgeries, these requests will only be processed under the following circumstances:
-1. Environment variable ```AWS_PROCESS_BEANSTALK_WORKER_REQUESTS``` is present and set to true (this should only be set in the worker environment).
-2. If the middleware is present, the request must originate from localhost or the Docker host.
+To add the middleware on application startup, set the ```AWS_PROCESS_BEANSTALK_WORKER_REQUESTS``` environment variable to true
+in the worker environment configuration.
+
+To protect against forgeries, daemon requests will only be processed if they originate from localhost or the Docker host.
 
 Periodic (scheduled) jobs are also supported with this approach without requiring any additional dependencies.
 Elastic Beanstalk workers support the addition of a ```cron.yaml``` file in the application root to configure this.
--- a/README.md
+++ b/README.md
@@ -355,6 +355,31 @@ require_relative 'config/environment' # load rails
 # To use this file, set CMD:  app.Aws::Rails::SqsActiveJob.lambda_job_handler
 ```
 
+### Elastic Beanstalk workers: processing activejobs using worker environments
+
+Another option for processing jobs without managing the worker process is hosting the application in a scalable 
+[Elastic Beanstalk worker environment](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features-managing-env-tiers.html). 
+This SDK includes Rack middleware that can be added conditionally and which will process requests from the 
+SQS Daemon provided with each worker instance. The middleware will forward each request and parameters to their appropriate jobs.
+
+To protect against forgeries, these requests will only be processed under the following circumstances:
+1. Environment variable ```AWS_PROCESS_BEANSTALK_WORKER_REQUESTS``` is present and set to true (this should only be set in the worker environment).
+2. If the middleware is present, the request must originate from localhost or the Docker host.
+
+Periodic (scheduled) jobs are also supported with this approach without requiring any additional dependencies.
+Elastic Beanstalk workers support the addition of a ```cron.yaml``` file in the application root to configure this.
+
+Example:
+```yml
+version: 1
+cron:
+ - name: "MyApplicationJob"
+   url: "/"
+   schedule: "0 */12 * * *"
+```
+
+Where 'name' must be the case-sensitive class name of the job. 
+
 ### Configuration
 
 For a complete list of configuration options see the
--- a/lib/aws/rails/middleware/elastic_beanstalk_worker_listener.rb
+++ b/lib/aws/rails/middleware/elastic_beanstalk_worker_listener.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+module Aws
+  module Rails
+    # Middleware to handle requests from the SQS Daemon present on Elastic Beanstalk worker environments.
+    class ElasticBeanstalkWorkerListener
+      INTERNAL_ERROR_MESSAGE = 'Failed to execute job - see Rails log for more details.'
+      INTERNAL_ERROR_RESPONSE = [500, { 'Content-Type' => 'text/plain' }, [INTERNAL_ERROR_MESSAGE]].freeze
+      FORBIDDEN_MESSAGE = 'Request with aws-sqsd user agent was made from untrusted address.'
+      FORBIDDEN_RESPONSE = [403, { 'Content-Type' => 'text/plain' }, [FORBIDDEN_MESSAGE]].freeze
+
+      def initialize(app)
+        @app = app
+        @logger = ActiveSupport::Logger.new(STDOUT)
+      end
+
+      def call(env)
+        request = ActionDispatch::Request.new env
+        if from_sqs_daemon?(request)
+          @logger.debug 'aws-rails-sdk middleware detected call from Elastic Beanstalk SQS Daemon.'
+
+          # Only accept requests from this user agent if it is from localhost or a docker host in case of forgery.
+          if request.local? || sent_from_docker_host?(request)
+            return periodic_task?(request) ? execute_periodic_task(request) : execute_job(request)
+          else
+            @logger.warn "SQSD request detected from untrusted address #{request.remote_ip}; returning 403 forbidden."
+            return FORBIDDEN_RESPONSE
+          end
+        end
+        @app.call(env)
+      end
+
+      private
+
+      def execute_job(request)
+        # Jobs queued from the Active Job SQS adapter contain the JSON message in the request body.
+        job = JSON.parse(request.body.string)
+        job_name = job['job_class']
+        @logger.debug "Executing job: #{job_name}"
+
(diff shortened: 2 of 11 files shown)


mullermp:
Thanks for opening a PR! We will review this shortly.

alextwoods:
Thanks for opening!  I'll take a look today.

I do agree with the decision not to implement the additional check against the Rails secret key - I think the restricted sending permissions should be sufficent, but it is an interesting idea and something we may want to add at some point (though as you point out, it wouldn't be backwards compatible and would need to be added as an opt-in/optional configuration).

lib/aws/rails/middleware/elastic_beanstalk_worker_listener.rb
```diff
@@ -0,0 +1,90 @@
...
+          end
+        end
+        @app.call(env)
+      end
+
+      private
+
+      def execute_job(request)
+        # Jobs queued from the Active Job SQS adapter contain the JSON message in the request body.
+        job = JSON.parse(request.body.string)
```
alextwoods, line 37: I'd recommend using the [JobRunner](https://github.com/aws/aws-sdk-rails/blob/master/lib/aws/rails/sqs_active_job/job_runner.rb) here - it does functionally do the same thing, but lets us keep that logic abstracted.  It also uses the Aws::Json (which lets users use whichever json engine they have configured rather).

lib/aws/rails/railtie.rb
```diff
@@ -64,5 +68,21 @@ def self.instrument_sdk_operations
         end
       end
     end
+
+    # Register a middleware that will handle requests from the Elastic Beanstalk worker SQS Daemon.
+    # This will only be added in the presence of the AWS_PROCESS_BEANSTALK_WORKER_REQUESTS environment variable.
+    # The expectation is this variable should only be set on EB worker environments.
+    def self.add_sqsd_listener_middleware(app)
+      is_eb_worker_hosted = ENV['AWS_PROCESS_BEANSTALK_WORKER_REQUESTS']
```
alextwoods, line 76: You can use the `Aws::Util.str_2_bool` method for this.

lib/aws/rails/middleware/elastic_beanstalk_worker_listener.rb
```diff
@@ -0,0 +1,90 @@
+# frozen_string_literal: true
+
+module Aws
+  module Rails
+    # Middleware to handle requests from the SQS Daemon present on Elastic Beanstalk worker environments.
+    class ElasticBeanstalkWorkerListener
```
alextwoods, line 6: I'd argue this isn't really a Listener -  maybe `EbsSqsActiveJobMiddleware` or something like that - trying to keep the name from being too long while being descriptive.  This more specific than handling SQS Daemon requests - it only handles requests that are translatable to ActiveJob (cron or sqs).

test/aws/rails/middleware/elastic_beanstalk_worker_listener_test.rb
```diff
@@ -0,0 +1,95 @@
...
+
+        test_middleware = ElasticBeanstalkWorkerListener.new(mock_rack_app)
+        response = test_middleware.call(mock_rack_env)
+
+        expect(response[0]).to eq(403)
+      end
+
+      it 'successfully invokes job when passed through request body' do
+        # Stub execute call to avoid invoking Active Job callbacks
+        allow(ActiveJob::Base).to receive(:execute).and_return(nil)
```
alextwoods, line 34: can change this from allow to expect since this is specifically what we are testing for here.

test/aws/rails/middleware/elastic_beanstalk_worker_listener_test.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      it 'successfully invokes periodic task when passed through custom header' do
+        mock_rack_env = create_mock_env('127.0.0.1', 'aws-sqsd/1.1', true)
+
+        test_middleware = ElasticBeanstalkWorkerListener.new(mock_rack_app)
+        response = test_middleware.call(mock_rack_env)
+
+        expect(response[0]).to eq(200)
+        expect(response[1]['Content-Type']).to eq('text/plain')
+        expect(response[2]).to eq(['Successfully ran periodic task ElasticBeanstalkPeriodicTask.'])
```
alextwoods, line 65: can we add an expectation that the `perform_now` method gets called?  Something like `expect_any_instance_of(ElasticBeanstalkPeriodicTask).to receive(:perform_now)`

test/aws/rails/railtie_test.rb
```diff
@@ -47,6 +47,55 @@ module Rails
         end
       end
 
+      describe '.add_sqsd_listener_middleware' do
+        it 'adds middleware when AWS_PROCESS_BEANSTALK_WORKER_REQUESTS is set to true' do
+          ENV['AWS_PROCESS_BEANSTALK_WORKER_REQUESTS'] = 'True'
```
alextwoods, line 52: Setting the ENV in a test could cause side effects, so if possible would be better to set in a before and unset in an after.  Probably wouldn't ever be an issue here since you're deleting it later in the test - but if the test fails/raises an exception before its deleted it will remain set.

alextwoods (commented):
Thanks for opening this - looks great: well structured, well tested and even well documented!  Just a few minor comments.

test/aws/rails/railtie_test.rb
mullermp: An RSpec `after` hook might work

README.md
```diff
@@ -355,6 +355,31 @@ require_relative 'config/environment' # load rails
...
 
+### Elastic Beanstalk workers: processing activejobs using worker environments
+
+Another option for processing jobs without managing the worker process is hosting the application in a scalable 
+[Elastic Beanstalk worker environment](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features-managing-env-tiers.html). 
+This SDK includes Rack middleware that can be added conditionally and which will process requests from the 
+SQS Daemon provided with each worker instance. The middleware will forward each request and parameters to their appropriate jobs.
+
+To protect against forgeries, these requests will only be processed under the following circumstances:
+1. Environment variable ```AWS_PROCESS_BEANSTALK_WORKER_REQUESTS``` is present and set to true (this should only be set in the worker environment).
```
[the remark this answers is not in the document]
MrWellington: Technically yes - my intent here was to highlight that the middleware can and should be omitted from web environments open to requests from outside. I'll do some rephrasing here.

lib/aws/rails/middleware/elastic_beanstalk_worker_listener.rb
MrWellington: Ah yeah makes sense - completely missed that class while working on this! Will update.

lib/aws/rails/railtie.rb
MrWellington: Sounds good - will do.

lib/aws/rails/middleware/elastic_beanstalk_worker_listener.rb
MrWellington: Yeah I didn't care for the name and was waffling on it...I like your proposed one better.

test/aws/rails/middleware/elastic_beanstalk_worker_listener_test.rb
MrWellington: Makes sense - first time working with RSpec and still getting a sense of its capabilities. Will need to read up on their docs more.

test/aws/rails/middleware/elastic_beanstalk_worker_listener_test.rb
MrWellington: Didn't realize you could do that! Naturally that's what I wanted to validate, I'll add the assertion.

test/aws/rails/railtie_test.rb
MrWellington: Good point on the exception - using an RSpec hook makes more sense.

MrWellington:
Just pushed in some changes addressing the comments above as well as a few stylistic updates recommended by a colleague. 

Only thing to note is I opted for a different change than the JobRunner approach - that class' constructor appears to expect a SQS message object. Given we're not actually polling SQS directly here and the daemon doesn't send all the parameters required to instantiate a message, didn't think it made sense to use here.

Alternatively could consider refactoring JobRunner and its consumers to just supply/take the job's JSON string as it doesn't currently require other fields in that object. In my case wanted to avoid impacting the existing code base as much as possible though.

lib/aws/rails/middleware/elastic_beanstalk_worker_listener.rb
alextwoods: Agree with your comments lower down on not using the JobRunner here since it takes the sqs message -  your update here looks good.

alextwoods (approved):
This looks good!  Last thing to do is just add a CHANGELOG.md entry (just add a `* Feature - ` entry under `Unreleased Changes`) and I can then merge this and release as version 3.6.0 of the gem.

MrWellington:
Just pushed in the update. Thanks everyone - appreciate the responsiveness and collaboration!

alextwoods:
Thanks again for the contribution!  I've merged the changes and they have been released as [aws-sdk-rails 3.6.0](https://rubygems.org/gems/aws-sdk-rails/versions/3.6.0)!

PR #74, opened 2021-09-13
SESv2 support

rromanchuk:
*Issue #, if available:*

*Description of changes:*


lib/aws/rails/railtie.rb
```diff
@@ -32,7 +32,8 @@ class Railtie < ::Rails::Railtie
     #   Aws::SES::Client initialization method.
     def self.add_action_mailer_delivery_method(name = :ses, options = {})
       ActiveSupport.on_load(:action_mailer) do
-        add_delivery_method(name, Aws::Rails::Mailer, options)
+        add_delivery_method(name, Aws::Rails::SesMailer, options)
+        add_delivery_method(:sesv2, Aws::Rails::Sesv2Mailer, options)
```
mullermp, line 36: I'm not a fan of this.. the name cannot be set for sesv2, and the options are the same. I think we'll need a second method and maybe a major version bump for this - or drop in replacement.

mullermp (commented):
Thanks for opening a pull request and implementing a first draft! I am wondering, can SESv2 be a drop in replacement for V1? Instead of having two mailers, replace v1 with v2? It's an internal detail from the API perspective.

lib/aws/rails/railtie.rb
rromanchuk: I'm running this on production, but really just submitting for visibility, for now at least, (and for myself when i forget in 2 weeks), especially since the documentation/implementation is sparse. I 100% offloaded subscription management, topic management, and deliverability (transactions/marketing) and just a raced to ship.  Literally didn't even run the tests 😆 

There are basically three interfaces to the same internal product. 

SESv1
Pinpoint
SESv2 which is taking the feature set of pinpoint, but making it not for aliens. It's WIP, superset of v1, just signatures/payloads aligning with the ability to manage deliverability/monitoring (config sets),  and subscription/list management. I like it, i see the road map, without actually knowing. 

 For any SEO travelers, here's boilerplate example,  in context of ActionMailer 

```ruby
class MarketingMailer < ApplicationMailer
  after_action :set_config_list_headers
  default from: PROMOTIONS_EMAIL

  def test
    mail subject: "Test marketing email"
  end

  # https://docs.aws.amazon.com/ses/latest/dg/sending-email-list-management.html
  # https://docs.aws.amazon.com/ses/latest/dg/using-configuration-sets-in-email.html
  def set_config_list_headers
    headers["X-SES-CONFIGURATION-SET"] = MARKETING_CONFIG_SET
    headers["X-SES-LIST-MANAGEMENT-OPTIONS"] = "myapp; topic=promotions"
  end
end
```

```erb
<%# _footer.html.mjml %>
<mj-section background-repeat="repeat" padding="20px 0" text-align="center">
  <mj-column>
    <mj-text align="center" color="#fff" font-size="13px" line-height="22px" padding="10px 25px">Copyright © 2021 blah blah</mj-text>

    <mj-text align="center" color="#fff" font-family="Arial, sans-serif" font-size="13px" line-height="22px" padding="10px 25px">
      <%# https://docs.aws.amazon.com/ses/latest/dg/sending-email-subscription-management.html %>
      <%= link_to "Unsubscribe", "{{amazonSESUnsubscribeUrl}}", style: "color:#535353", "ses:no-track": "" %>
    </mj-text>
  </mj-column>
</mj-section>
```

rromanchuk: @mullermp I'm still using this in production, so I should come back and look at this in context of the sdk instead of proof of concept.  I'm guessing replacement and version bump is best route and should just feel like a superset. Just need to sit down and compare to v1 for compatibility issues and clean it up.

mullermp: Hey - thanks for following up. I'm not against this change - it just needs clean up and needs to address backwards compatibility concerns.

To summarize:

I think adding aws-sdk-sesv2 as a dependency should be safe in a minor version.

If we do this in the current MV, we would probably need to keep `Aws::Rails::Mailer = Aws::Rails::SesMailer ` but we should also add a comment for why.

We might be able to add another positional parameter to `add_action_mailer_delivery_method` for enabling ses_v2. Something like this:

```
    # rename options to client_options for clarity
    # add mailer param with default of :ses, but can take :ses_v2
    def self.add_action_mailer_delivery_method(name = :ses, client_options = {}, mailer = :ses)
      ActiveSupport.on_load(:action_mailer) do
        case mailer
        when :ses
          add_delivery_method(name, Aws::Rails::SesMailer, options)
        when :sesv2
          # also still name
          add_delivery_method(name, Aws::Rails::SesV2Mailer, options)
        else
          # figure out what to do here, but probably raise an ArgumentError
        end
      end
    end
```

ssinghi: can we perhaps make this change, and make a release with experimental SESv2 support? 
Let me know if you want me to jump on this and update the PR, or create a new PR? 
Thanks!

mullermp: Thanks ssinghi. Our team has other initiatives going on so I haven't been able to work on this. If you'd like to move it forward, I can assist in review and release.

mullermp:
Closing in favor of more recent PR #93

PR #77, opened 2022-01-17
Conform with Rails session store requirements

### LINKED ISSUE #76: DynamoDB Session Store does not work with Rails 7
DynamoDB Session Store (dynamodb_store) does not work with Rails 7.

The most simple setup (a fresh Rails 7.0.1 project with an empty action) would generate the following error:

```
NoMethodError (undefined method `enabled?' for {"_csrf_token"=>"..."}:Rack::Session::Abstract::SessionHash):

actionpack (7.0.1) lib/action_dispatch/middleware/flash.rb:62:in `commit_flash'
actionpack (7.0.1) lib/action_controller/metal.rb:189:in `dispatch'
actionpack (7.0.1) lib/action_controller/metal.rb:251:in `dispatch'
actionpack (7.0.1) lib/action_dispatch/routing/route_set.rb:49:in `dispatch'
actionpack (7.0.1) lib/action_dispatch/routing/route_set.rb:32:in `serve'
actionpack (7.0.1) lib/action_dispatch/journey/router.rb:50:in `block in serve'
actionpack (7.0.1) lib/action_dispatch/journey/router.rb:32:in `each'
actionpack (7.0.1) lib/action_dispatch/journey/router.rb:32:in `serve'
actionpack (7.0.1) lib/action_dispatch/routing/route_set.rb:850:in `call'
(snip)
```

The direct trigger is this change https://github.com/rails/rails/commit/ca7c820c6962788d510c99a68e725de83cbe9cad which accesses [`session.enabled?`](https://github.com/rails/rails/blob/v7.0.1/actionpack/lib/action_dispatch/request/session.rb#L87-L89) in the request dispatching.

However, the root cause seems to lie in dynamodb_store's implementation. When dynamodb_store is used, the `session` variable is a Rack::Session::Abstract::SessionHash, where it should be a ActionDispatch::Request::Session. `session.enabled?` is implemented only in ActionDispatch::Request::Session, thus resulting in the NoMethodError.

osyoyu:
Fixes #76.

Include ~three~ two modules which are included in ActionDispatch::Session::AbstractStore (which is not inherited in the current implementation, at least as for now).

SessionStore is especially important; it overrides configuration of Rack headers and makes the request-scope `session` object an ActionDispatch::Request::Session object, matching Rails' expectations.

It may be noted that this fix is not optimal in terms of robustness; it might break once ActionDispatch::Session::AbstractStore changes its internal implementations.
Changing [Aws::SessionStore::DynamoDB::RackMiddleware](https://github.com/aws/aws-sessionstore-dynamodb-ruby/blob/v2.0.1/lib/aws/session_store/dynamo_db/rack_middleware.rb)'s superclass to ActionDispatch::Session::AbstractStore would be a future-proof way. However, that will make aws-sessionstore-dynamodb-ruby depend on Rails, breaking compatibility for non-Rails users.
This patch simply chooses the safer way.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Fix DynamoDB session store to work with Rails 7.
+
 3.6.1 (2021-06-08)
 ------------------
 
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,11 @@ group :test do
   gem 'rspec-expectations'
   gem 'rspec-mocks'
   gem 'bcrypt'
+  if defined?(JRUBY_VERSION)
+    gem 'activerecord-jdbcsqlite3-adapter'
+  else
+    gem 'sqlite3'
+  end
 end
 
 group :docs do
--- a/Gemfile
+++ b/Gemfile
@@ -12,11 +12,6 @@ group :test do
   gem 'rspec-expectations'
   gem 'rspec-mocks'
   gem 'bcrypt'
-  if defined?(JRUBY_VERSION)
-    gem 'activerecord-jdbcsqlite3-adapter'
-  else
-    gem 'sqlite3'
-  end
 end
 
 group :docs do
--- a/Rakefile
+++ b/Rakefile
@@ -11,6 +11,7 @@ Dir.glob('**/*.rake').each do |task_file|
 end
 
 Rake::TestTask.new do |t|
+  ENV['DATABASE_URL'] = 'sqlite3::memory:'
   t.libs << 'test'
   t.pattern = 'test/**/*_test.rb'
   t.warning = false
--- a/Rakefile
+++ b/Rakefile
@@ -11,7 +11,6 @@ Dir.glob('**/*.rake').each do |task_file|
 end
 
 Rake::TestTask.new do |t|
-  ENV['DATABASE_URL'] = 'sqlite3::memory:'
   t.libs << 'test'
   t.pattern = 'test/**/*_test.rb'
   t.warning = false
--- a/aws-sdk-rails.gemspec
+++ b/aws-sdk-rails.gemspec
@@ -25,4 +25,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency('railties', '>= 5.2.0') # encrypted credentials
   spec.add_dependency('concurrent-ruby', '~> 1') # Utilities for concurrent processing
   spec.add_development_dependency('rails')
+  spec.add_development_dependency('sqlite3')
 end
--- a/aws-sdk-rails.gemspec
+++ b/aws-sdk-rails.gemspec
@@ -25,5 +25,4 @@ Gem::Specification.new do |spec|
   spec.add_dependency('railties', '>= 5.2.0') # encrypted credentials
   spec.add_dependency('concurrent-ruby', '~> 1') # Utilities for concurrent processing
   spec.add_development_dependency('rails')
-  spec.add_development_dependency('sqlite3')
 end
--- a/lib/action_dispatch/session/dynamodb_store.rb
+++ b/lib/action_dispatch/session/dynamodb_store.rb
@@ -15,7 +15,6 @@ module ActionDispatch
     #
     # @see https://docs.aws.amazon.com/sdk-for-ruby/aws-sessionstore-dynamodb/api/Aws/SessionStore/DynamoDB/Configuration.html
     class DynamodbStore < Aws::SessionStore::DynamoDB::RackMiddleware
-      include Compatibility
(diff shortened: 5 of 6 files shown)


mullermp:
Thanks for opening this pull request. I think your approach makes sense. Rails main gemfile tests appear to be failing. It looks unrelated(?) but perhaps we should fix it for completeness.

osyoyu:
Oops, sorry, didn't notice tests were failing in local.
Not loading Active Record fixes this problem. Should I include the following patch in this pull request, or just submit another one?

```diff
diff --git a/test/dummy/config/application.rb b/test/dummy/config/application.rb
index 1737fff..e4b29ed 100644
--- a/test/dummy/config/application.rb
+++ b/test/dummy/config/application.rb
@@ -1,6 +1,19 @@
 # frozen_string_literal: true

-require 'rails/all'
+require 'rails'
+# require 'active_model/railtie'
+require 'active_job/railtie'
+# require 'active_record/railtie'
+# require 'active_storage/engine'
+require 'action_controller/railtie'
+require 'action_mailer/railtie'
+# require 'action_mailbox/engine'
+# require 'action_text/engine'
+# require 'action_view/railtie'
+# require 'action_cable/engine'
+# require 'sprockets/railtie'
+# require 'rails/test_unit/railtie'
+
 require 'aws-sdk-rails'

 module Dummy
```

mullermp:
You can include changes in this PR, that's fine. When writing that patch, did you happen to see what change Rails made to break this?

mullermp:
I'd like to keep rails/all for the test app if possible.

osyoyu:
A bisect found out https://github.com/rails/rails/commit/265c1e9f2d9e8b4747661ba66e2096196257505f to be the cause.

> I'd like to keep rails/all for the test app if possible.

Sure thing. I defined a DATABASE_URL that points to an in-memory SQLite database (that adds a dependency to sqlite3 gem in test environments). Now tests should run fine.

osyoyu:
9ffd07b7cbbe74f0daf9fe0a4977bec44942b28c removes the [Compatibility module](https://github.com/rails/rails/blob/v7.0.1/actionpack/lib/action_dispatch/middleware/session/abstract_store.rb#L20) which is included in ActionDispatch::Session::AbstractStore. This is because Compatibility#generate_sid is incompatible with [DynamodbStore's version of generate_sid](https://github.com/aws/aws-sessionstore-dynamodb-ruby/blob/v2.0.1/lib/aws/session_store/dynamo_db/rack_middleware.rb#L103-L107), which makes [sid verification](https://github.com/aws/aws-sessionstore-dynamodb-ruby/blob/v2.0.1/lib/aws/session_store/dynamo_db/rack_middleware.rb#L53) always fail.

I'm not sure if any easy fix exists (besides of removing Compatibility), as long as RackMiddleware inherits Rack::Session::Abstract::Persisted instead of  ActionDispatch::Session::AbstractStore.

osyoyu:
and sqlite3 is incompatible with JRuby, of course...

mullermp:
I appreciate you digging into this... @alextwoods is also taking a look.

mullermp:
Does sqlite3 need to be installed, or can we simply just set the database URL to something nonsensical? Otherwise, I believe for JRuby, you need to use the jdbc sqlite3 gem.

osyoyu:
An adapter matching the scheme part in DATABASE_URL must be installed -- something like `foo://bar` would raise an exception.
I modified the Gemfile so that activerecord-jdbcsqlite3-adapter will be used in JRuby. Hope this works

mullermp (approved):

osyoyu:
Tests passed! :sun:

Another note: Including the Compatibility module would have introduced another incompatibility 😁 The default cookie name would have changed from `rack.session` to `_session_id`, leaving users with invalidated sessions.
https://github.com/rails/rails/blob/v7.0.1/actionpack/lib/action_dispatch/middleware/session/abstract_store.rb#L22

mullermp:
>Compatibility module would have introduced another incompatibility
Haaaa... great catch. I think it makes sense to leave it out.

Thanks again for your contributions. Before we release this, we will need to do some manual tests. We have a sample_app in the root of this repo. We will aim to get this released this week. 👍

alextwoods:
Thanks again for the contribution and the work debugging the tests - I pushed a small tweak to avoid needing to set the db url or use sqllite.

alextwoods (approved):

mullermp:
🤦 I wanted to keep rails/all

> I'd like to keep rails/all for the test app if possible.

mullermp:
Let's just keep it as is.. Sorry for the churn.

osyoyu:
That's ok 👍 Thanks for your swift review!

PR #79, opened 2022-01-20
Upgrade sample app to rails 7

alextwoods:
*Description of changes:*

### DIFF
--- a/sample_app/Gemfile
+++ b/sample_app/Gemfile
@@ -7,7 +7,11 @@ gem 'aws-sdk-rails', path: '../'
 # gem 'aws-sessionstore-dynamodb', path: '../../aws-sessionstore-dynamodb-ruby'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails' #, path: '../../rails'
+gem 'rails', '~> 7.0'
+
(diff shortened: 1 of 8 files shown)


sample_app/Gemfile
```diff
@@ -7,7 +7,11 @@ gem 'aws-sdk-rails', path: '../'
 # gem 'aws-sessionstore-dynamodb', path: '../../aws-sessionstore-dynamodb-ruby'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails' #, path: '../../rails'
+gem 'rails', '~> 7.0'
```
mullermp, line 10: I prefer to keep the comment (for development)

mullermp (approved):
Assuming most of these are "generated" changes from rails new?

sample_app/Gemfile
alextwoods: These are mostly changes from running a `rails app:upgrade` - we could instead create a new app (i think in particular rails 7 does not use webpack by default) and add our customizations in instead so we get a cleaner rails 7 app.  Thoughts?

mullermp: I didn't know about that command. I'm sure it's fine.

PR #81, opened 2022-06-14
Dynamic message group ids

rdubya:
- This fixes the existing failing tests in Ruby 3 that appear to be due to keyword arguments.
- This adds the ability to define a `message_group_id` method on the job that will be called if using a fifo queue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,12 +420,18 @@ If the order in which your jobs executes is important, consider using a
 A FIFO queue ensures that messages are processed in the order they were sent
 (First-In-First-Out) and exactly-once processing (ensuring duplicates are never
 introduced into the queue).  To use a fifo queue, simply set the queue url (which will end in ".fifo")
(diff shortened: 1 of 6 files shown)


alextwoods (commented):
Thanks for the submission - the change over all looks good! Could you add an entry to the CHANGELOG in the unreleased section (as an issue) and we can then release a new version.

rdubya:
Great! I added a change log entry. Let me know if there is anything else you would like to see.

alextwoods (approved):
Looks great! Thanks again for the contribution.  I'll merge this and release a new version of the gem.

alextwoods:
Released as 3.6.2: https://rubygems.org/gems/aws-sdk-rails/versions/3.6.2

rdubya:
Awesome. Thanks for the quick turnaround.

PR #82, opened 2022-06-16
Removed superfluous to_sym.

tetsuya-ogawa:
*Issue #, if available:*

Actually, I wanted queue_url_for to work even if the queues key is a string.

But I thought it would not be elegant to do that in a gem, so I'll just give a PR for the parts I'm interested in.

*Description of changes:*

Removed superfluous to_sym.


alextwoods:
The change itself looks reasonable here - I did merge main into your branch to fix the failing ruby 3 tests, which are passing now.

I'm curious though about your use case and how removing this helps with what you want to do?

tetsuya-ogawa:
Thanks.

There is no purpose to this change. I just removed it because I didn't think I needed it.
Additionally, I wanted to commit to a gem that I use all the time. ：)

PR #85, opened 2022-08-24
Support i18n on jobs enqueued via amazon_sqs_async_adapter

nov:
*Issue #, if available:*

*Description of changes:*

`amazon_sqs_async_adapter` loses `I18n.locale` because of `Concurrency::Promise`

to pass `I18n.locale` to promise, use `Concurrency::Promises.future` instead of `Concurrency::Promise.execute`


mullermp:
Thanks for opening a pull request. This change seems fine. Do you have any other information about this issue, and why Promise doesn't work but future does?

nov:
Hi, I'm not so familiar with `concurrent-ruby`, but from its README, `Concurrency::Promise` is deprecated and `Concurrency::Promises` is recommended as the alternative.

However, I'm OK with `Concurrency::Promise` if it can pass main thread status to the child too.

mullermp:
Ah - ok! Then yes, we should move to `Promises`. We're happy to take this change. I would like to see a test case for I18n being used as expected.

mullermp:
I've added a few tests for I18n. In your code, you added `defined?(I18n)`. My understanding is that I18n is always available in Rails? Curious to know if you think otherwise. I suppose in future versions of Rails it *could* be removed though unlikely.

mullermp:
Before merging, I'm doing some investigation. I just want to make sure this is the correct change to make.

I'm seeing in ActiveJob that locale is already serialized onto the message - https://github.com/rails/rails/blob/main/activejob/lib/active_job/core.rb#L114 . Your job, whether sync or async, should be using that locale!

It's possible that your locale is dropped not because of Promise but because "I18n.locale can leak into subsequent requests served by the same thread/process if it is not consistently set in every controller." - https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests . I'm also not seeing it handled by ActiveJob here - https://github.com/rails/rails/blob/04972d9b9ef60796dc8f0917817b5392d61fcf09/activejob/lib/active_job/queue_adapters/async_adapter.rb

mullermp:
I think this approach makes more sense. This one was very obscure. We found that jobs are already wrapped with I18n, but it worked with Rails async adapter because they serialize the job prior to passing it to the thread executor. So we take the same approach here - serialize the job (calls I18n.locale at that moment) then pass to Concurrent. I also added some other clean ups.

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
```diff
@@ -11,16 +11,16 @@ def enqueue(job)
         _enqueue(job)
       end
 
-      def enqueue_at(job, timestamp, opts={})
+      def enqueue_at(job, timestamp)
         delay = (timestamp - Time.now.to_f).floor
         raise ArgumentError, 'Unable to queue a job with a delay great than 15 minutes' if delay > 15.minutes
-        _enqueue(job, delay_seconds: delay)
+        _enqueue(job, {}, delay_seconds: delay)
```
alextwoods, line 17: I dont think this is correct - should probably be nil (or job.serialize)

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
```diff
@@ -11,16 +11,16 @@ def enqueue(job)
...
       end
 
       private
 
-      def _enqueue(job, send_message_opts = {})
-        body = job.serialize
+      def _enqueue(job, body = nil, send_message_opts = {})
```
alextwoods, line 22: Maybe body can be moved into the opts here and deleted? saves us having to set it to nil elsewhere when we dont want this.  and then just:
```
body ||= send_message_opts.delete(:body) || job.serialize
```

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
mullermp: Duplicate comment, but fixed.

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
mullermp: I think it's fine to pass nil here because it's a private method and only need to do it once. It's easier to understand this way. If I were to put it in send_message_opts, we should just require it to be json dumped by the caller too.

alextwoods (approved):

alextwoods (approved):
LGTM - could use a changelog

nov:
In Rails app, `I18n` is always available.
The reason I put `defined?(I18n)` is just because this gem itself doesn't have `I18n` dependency and I was not sure I should add it as `development_dependency`.

And serializing job before passing it to the thread is much better than my code ;)

nov:
hi, is there any plan to release new version of this gem?

mullermp:
I'm so sorry. This slipped my mind. We will do this today.

mullermp:
This should be released now as 3.6.3

nov:
thanks a lot!

PR #87, opened 2022-09-20
Enable automatic message visibility refresh

notac:
*Description of changes:*

This PR will enable the auto-refresh of the visibility timeout of active messages.

This will allow users to use much lower default visibility timeouts, which will ensure messages can be returned to the queue promptly after a job fails, etc.

NOTE: I didn't do the work of adding automated tests yet, will do so if the work is acceptable.

### DIFF
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -52,10 +52,9 @@ module Aws
                 # Extend the visibility timeout to the provided refresh timeout
                 message.change_visibility({ visibility_timeout: @refresh_timeout })
                 # Wait half the refresh timeout, and repeat
-                sleep(@refresh_timeout / 2)
+                sleep(@refresh_timeout / 2.0)
               rescue => e
                 # If anything goes wrong, we want to handle it. We don't care what.
-                @logger.error("Monitor process failed for message: #{message.id}. Error: #{e}")
               end
             end
           end if @monitor
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -43,7 +43,7 @@ module Aws
               @logger.info "Error processing job #{job_msg}: #{e}"
               @logger.debug e.backtrace.join("\n")
             ensure
-              refresh_visibility = true
+              refresh_visibility = false
             end
           end
 
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -54,10 +54,10 @@ module Aws
                 message.change_visibility({ visibility_timeout: @visibility_refresh })
                 # Wait half the refresh timeout, and repeat
                 sleep(@visibility_refresh / 2.0)
-              rescue Aws::SQS::Errors::ReceiptHandleIsInvalid =>
+              rescue Aws::SQS::Errors::ReceiptHandleIsInvalid
                 # Message has been deleted. We don't care!
                 break
-              rescue Aws::SQS::Errors::MessageNotInflight => e
+              rescue Aws::SQS::Errors::MessageNotInflight
                 # Message has been otherwise released. We don't care!
                 break
               rescue => e
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -27,7 +27,7 @@ module Aws
         # TODO: Consider catching the exception and sleeping instead of using :caller_runs
         def execute(message)
           # Used to tell the visibilty refresh thread to give up
-          poison_pill = false
+          refresh_visibility = true
           @executor.post(message) do |message|
             begin
               job = JobRunner.new(message)
@@ -42,12 +42,12 @@ module Aws
               @logger.info "Error processing job #{job_msg}: #{e}"
               @logger.debug e.backtrace.join("\n")
             ensure
-              poison_pill = true
+              refresh_visibility = true
             end
           end
 
           @monitor.post(message) do |message|
-            while poison_pill
+            while refresh_visibility
               begin
                 # Extend the visibility timeout to the provided refresh timeout
                 message.change_visibility({ visibility_timeout: @refresh_timeout })
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -17,18 +17,18 @@ module Aws
            fallback_policy: :caller_runs # slow down the producer thread
         }.freeze
 
-        def initialize(options = {}, refresh_timeout = nil)
+        def initialize(options = {})
           @executor = Concurrent::ThreadPoolExecutor.new(DEFAULTS.merge(options))
           # Monitor threads used to refresh visiblity
-          @refresh_timeout = refresh_timeout
-          @monitor = Concurrent::ThreadPoolExecutor.new(DEFAULTS.merge(options)) if refresh_timeout
+          @visibility_refresh = options.delete(:visibility_refresh)
+          @monitor = Concurrent::ThreadPoolExecutor.new(DEFAULTS.merge(options)) if @visibility_refresh
           @logger = options[:logger] || ActiveSupport::Logger.new(STDOUT)
         end
 
         # TODO: Consider catching the exception and sleeping instead of using :caller_runs
         def execute(message)
           # Used to tell the visibilty refresh thread to give up
-          refresh_visibility = true
+          refresh_monitor = true
           @executor.post(message) do |message|
             begin
               job = JobRunner.new(message)
@@ -43,17 +43,17 @@ module Aws
               @logger.info "Error processing job #{job_msg}: #{e}"
               @logger.debug e.backtrace.join("\n")
             ensure
-              refresh_visibility = false
+              refresh_monitor = false
             end
           end
 
           @monitor.post(message) do |message|
-            while refresh_visibility
+            while refresh_monitor
               begin
                 # Extend the visibility timeout to the provided refresh timeout
-                message.change_visibility({ visibility_timeout: @refresh_timeout })
+                message.change_visibility({ visibility_timeout: @visibility_refresh })
                 # Wait half the refresh timeout, and repeat
-                sleep(@refresh_timeout / 2.0)
+                sleep(@visibility_refresh / 2.0)
               rescue Aws::SQS::Errors::MessageNotInflight => e
                 # Message has been deleted or otherwise released. We don't care!
                 break
@@ -63,7 +63,7 @@ module Aws
                 break
               end
             end
-          end if @monitor
+          end if @visibility_refresh
         end
 
         def shutdown(timeout=nil)
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -49,9 +49,9 @@ module Aws
           @monitor.post(message) do |message|
             while poison_pill
               begin
-                # Extend the visibility timeout by one minute
-                message.change_visibility({ visibility_timeout: 1.minute })
-                # Wait 30 seconds, and repeat
+                # Extend the visibility timeout to the provided refresh timeout
+                message.change_visibility({ visibility_timeout: @refresh_timeout })
+                # Wait half the refresh timeout, and repeat
                 sleep(@refresh_timeout / 2)
               rescue => e
                 # If anything goes wrong, we want to handle it. We don't care what.
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -46,7 +46,10 @@ module Aws
               refresh_monitor = false
             end
           end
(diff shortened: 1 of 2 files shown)


lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +41,24 @@ def execute(message)
...
             end
           end
+
+          @monitor.post(message) do |message|
+            while poison_pill
+              begin
+                # Extend the visibility timeout to the provided refresh timeout
+                message.change_visibility({ visibility_timeout: @refresh_timeout })
+                # Wait half the refresh timeout, and repeat
+                sleep(@refresh_timeout / 2)
```
alextwoods, line 55: Maybe use / 2.0 to prevent possible integer rounding issues.

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +41,24 @@ def execute(message)
...
+              rescue => e
+                # If anything goes wrong, we want to handle it. We don't care what.
```
alextwoods, line 57: We may want to not log certain issues - like if the message has already been deleted and the change_visibility call fails.  Haven't thought through all of the error cases, but at least that one seems expected.

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -16,13 +16,18 @@ class Executor
            fallback_policy: :caller_runs # slow down the producer thread
         }.freeze
 
-        def initialize(options = {})
+        def initialize(options = {}, refresh_timeout = nil)
```
alextwoods, line 19: Why not include the refresh_timeout in the options?  This leads to a bit of a clunky interface when setting both.

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -16,13 +16,18 @@ class Executor
...
+          # Monitor threads used to refresh visiblity
+          @refresh_timeout = refresh_timeout
+          @monitor = Concurrent::ThreadPoolExecutor.new(DEFAULTS.merge(options)) if refresh_timeout
           @logger = options[:logger] || ActiveSupport::Logger.new(STDOUT)
         end
 
         # TODO: Consider catching the exception and sleeping instead of using :caller_runs
         def execute(message)
+          # Used to tell the visibilty refresh thread to give up
+          poison_pill = false
```
alextwoods, line 30: The name `poison_pill` here is a little confusing  - and I'm not sure I understand the logic below in the `@monitor.post` block.  The monitor will update the visibility `while poison_pill`, but `poison_pill` starts as false and is only set to true once execution (successfully or unsuccessfully) is complete.  

Maybe consider updating this to be `job_executing` and set it to `true` initially, then `false` in the ensure block.

alextwoods (commented):
Thanks for the contribution! Overall approach I believe makes sense - this is a useful feature!

lib/aws/rails/sqs_active_job/executor.rb
notac: Was leaving it out since options is merged straight into defaults and passed to ThreadPoolExecutor. Happy to move it if you prefer combined.

lib/aws/rails/sqs_active_job/executor.rb
notac: My bad, flipped it around in my brain, due to bad naming 😆 Fixing.

lib/aws/rails/sqs_active_job/executor.rb
notac: I can leave this out and just drop out for now?

notac: Added the catch for that. I don't see any other exceptions that we should treat specially?

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +42,28 @@ def execute(message)
...
           end
+
+          @monitor.post(message) do |message|
+            while refresh_visibility
+              begin
+                # Extend the visibility timeout to the provided refresh timeout
+                message.change_visibility({ visibility_timeout: @refresh_timeout })
+                # Wait half the refresh timeout, and repeat
+                sleep(@refresh_timeout / 2.0)
+              rescue Aws::SQS::Errors::MessageNotInflight => e
```
notac, line 57: looks like I need to handle ```Aws::SQS::Errors::ReceiptHandleIsInvalid``` as well. Will add that in the a.m.

lib/aws/rails/sqs_active_job/executor.rb
mullermp: You can delete it out

alextwoods: Yeah, +1 to doing something like `@refresh_timeout = option.delete(:refresh_timeout)`

lib/aws/rails/sqs_active_job/poller.rb
```diff
@@ -112,6 +116,7 @@ def parse_args(argv)
             opts.on("-t", "--threads INTEGER", Integer, "The maximum number of worker threads to create.  Defaults to 2x the number of processors available on this system.") { |a| out[:threads] = a }
             opts.on("-b", "--backpressure INTEGER", Integer, "The maximum number of messages to have waiting in the Executor queue. This should be a low, but non zero number.  Messages in the Executor queue cannot be picked up by other processes and will slow down shutdown.") { |a| out[:backpressure] = a }
             opts.on("-m", "--max_messages INTEGER", Integer, "Max number of messages to receive in a batch from SQS.") { |a| out[:max_messages] = a }
+            opts.on("-r", "--refresh_timeout", Integer, "Enables message visibility refresh, Refresh timeout is in seconds.") { |a| out[:refresh_timeout] = a }
```
alextwoods, line 119: For parameter name, I think we should use something with `visibility` in the name to make it clear what this applies to.  Maybe `visibility_refresh`?

notac: Done

lib/aws/rails/sqs_active_job/executor.rb
notac: Done

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +42,31 @@ def execute(message)
               job_msg = job ? "#{job.id}[#{job.class_name}]" : 'unknown job'
               @logger.info "Error processing job #{job_msg}: #{e}"
               @logger.debug e.backtrace.join("\n")
+            ensure
+              refresh_monitor = false
             end
           end
+
+          @monitor.post(message) do |message|
+            while refresh_monitor
```
jshah, line 51: won't `refresh_monitor` always be `false` because of the `ensure`? We want to set it to `false` only when the message is successful?

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +42,31 @@ def execute(message)
...
+              rescue Aws::SQS::Errors::MessageNotInflight => e
+                # Message has been otherwise released. We don't care!
+                break
+              rescue => e
+                # If anything else goes wrong, we should log and break the loop.
+                @logger.error("Monitor process failed for message: #{message.id}. Error: #{e}")
+                break
+              end
+            end
+          end if @visibility_refresh
```
jshah, line 69: can we move this to the front of the method for readability?

```
if @visibility_refresh
   ...
end
```

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +42,31 @@ def execute(message)
...
+            ensure
+              refresh_monitor = false
             end
           end
+
+          @monitor.post(message) do |message|
+            while refresh_monitor
+              begin
+                # Extend the visibility timeout to the provided refresh timeout
+                message.change_visibility({ visibility_timeout: @visibility_refresh })
```
jshah, line 54: is it more accurate to say update the `visibility_timeout` so the job is available to be picked up again after failure?

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +42,31 @@ def execute(message)
...
+                # Wait half the refresh timeout, and repeat
+                sleep(@visibility_refresh / 2.0)
```
jshah, line 56: why do we want to sleep? if I set this to 6 hours, a worker would be stalled for a while.

On the other hand, If we only are in this block when a message fails, then just changing the visibility will allow it to get picked up by a worker again

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -18,11 +19,16 @@ class Executor
 
         def initialize(options = {})
           @executor = Concurrent::ThreadPoolExecutor.new(DEFAULTS.merge(options))
+          # Monitor threads used to refresh visiblity
+          @visibility_refresh = options.delete(:visibility_refresh)
+          @monitor = Concurrent::ThreadPoolExecutor.new(DEFAULTS.merge(options)) if @visibility_refresh
```
jshah, line 24: why do we need another thread here? can't we just use `@executor`, the logic is happening synchronously anyways

jshah (commented):
left a few comments to understand this approach, let me know if they make sense

jshah (commented):
We also want this refresh to happen during the `shutdown` method if the jobs are in the middle of being run.

See https://github.com/aws/aws-sdk-rails/blob/main/lib/aws/rails/sqs_active_job/executor.rb#L43-L53

lib/aws/rails/sqs_active_job/executor.rb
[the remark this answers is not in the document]
jshah: also, does the `@monitor` thread need to have `shutdown` logic and be added to https://github.com/aws/aws-sdk-rails/blob/main/lib/aws/rails/sqs_active_job/poller.rb#L62-L64?

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,8 +42,31 @@ def execute(message)
               job_msg = job ? "#{job.id}[#{job.class_name}]" : 'unknown job'
               @logger.info "Error processing job #{job_msg}: #{e}"
               @logger.debug e.backtrace.join("\n")
+            ensure
+              refresh_monitor = false
             end
           end
+
+          @monitor.post(message) do |message|
```
jshah, line 50: will this blow up if `@visibility_refresh` is not set since `@monitor` will be `nil`

lib/aws/rails/sqs_active_job/executor.rb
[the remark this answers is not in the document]
notac: If this is lower, what is submitted to change_visibility will be respected first.

lib/aws/rails/sqs_active_job/executor.rb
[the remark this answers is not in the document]
notac: Nah, as soon as the parent thread exits refresh_monitor will be set to false so this one will shut down. If a new job is picked up a new thread will be spun.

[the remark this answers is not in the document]
jshah: > Nah, as soon as the parent thread exits refresh_monitor will be set to false so this one will shut down

but when will it shut down? if this thread is sleeping, it could be shut down way after the parent thread finishes

lib/aws/rails/sqs_active_job/executor.rb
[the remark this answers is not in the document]
jshah: I would expect this to throw an error since 12 hours is the max. We can't extend past that.

Additionally, I'm thinking of this scenario:

1. `visiblity_timeout` is 2 hours, `visbility_refresh` is 1 hour
2. we receive a message, and start processing. `visibility_timeout` is extended to 3 hours
3. sleep for 30 minutes
4. 30 minutes elapses (30 mins total), `visibility_timeout` is extended to 4 hours
5. 30 minutes elapses (1 hour total), `visibility_timeout` is extended to 5 hours
6. ...

In the above scenario, won't we hit the limit much faster than the time the job has to run? Basically what I'm trying to say is that we should cover the case where the `visiblity_timeout` has already maxed out.

mullermp:
Thanks for making updates. It looks like tests are failing, likely syntax.

mullermp:
@notac We're looking to release a minor version of this gem soon with some other changes. I've fixed the syntax errors for this PR. This cannot be merged until we add some tests and address comments from @jshah. Were you still interested in moving this forward?

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -36,6 +42,32 @@ def execute(message)
...
               @logger.info "Error processing job #{job_msg}: #{e}"
               @logger.debug e.backtrace.join("\n")
+            ensure
+              refresh_monitor = false
+            end
+          end
+          refresh_monitor(message, refresh_monitor) if @visibility_refresh
+        end
+
+        def refresh_monitor(message, refresh_monitor)
```
SteffanPerry, line 52: the variable refresh_monitor is the same as the method name

mullermp:
This functionality has been moved to https://github.com/aws/aws-activejob-sqs-ruby Please open a new PR or issue there.

PR #91, opened 2022-11-18
Load aliases for Psych 4 (included in Ruby 3.1) 

ohbarye:
## Issue

N/A

## Description of changes

`YAML.load` is incompatible between Psych v4 and priors because Psych v4 has a breaking change.

ref: https://bugs.ruby-lang.org/issues/17866

To keep the current gem's behavior compatible between Psych v4 and priors (Ruby 3.1.0 and priors), it's necessary to use `YAML.unsafe_load` for Psych v4.

## Testing

All tests passed on my local with Ruby 3.1.2.
But I found that this repository's CI does not run tests on Ruby 3.1, so I'll try to add the version to [ci.yml](https://github.com/aws/aws-sdk-rails/blob/11c86231e2466d71854000d13162222fb50bec24/.github/workflows/ci.yml#L18).

---

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

PR #92, opened 2022-11-18
Add Ruby 3.1 to test targets and some improvements for CI

ohbarye:
## Issue

N/A

## Description of changes

While reading the codebase, I found it could have incompatibility with Ruby v3.1 (e.g. https://github.com/aws/aws-sdk-rails/pull/91) and this repository does not run tests with Ruby 3.1.

This pull request adds Ruby 3.1 to test targets and some improvements for CI.

- Bump actions/checkout to the latest version, v3: https://github.com/actions/checkout/blob/bf085276cecdb0cc76fbbe0687a5a0e786646936/CHANGELOG.md
- Enable `bundler-cache: true` to speed up CI
- Remove `.ruby-version` file
   - The file is added by https://github.com/aws/aws-sdk-rails/pull/85, I wasn't able to find the reason why.
   - IMO, `.ruby-version` is unnecessary for a gem that supports multiple Ruby versions like `aws-sdk-rails`.

## Question

Ruby 2.6 and below are already EOL. Should this gem continue to support them?

ref https://endoflife.date/ruby

---


alextwoods:
Looks like Ruby 3.1 is broken for Rails 6.0 and 6.1 - we may need to exclude those as well.

alextwoods:
While Ruby <2.6 and Rails 5.2 are EOL, I'd like to at least know when we make a change that would break them and make a decision at that point, so lets leave them in the CI for now.

As for removal of .ruby-version - good catch, I think it was an accidental add in that PR.

ohbarye:
@alextwoods 
>Looks like Ruby 3.1 is broken for Rails 6.0 and 6.1 - we may need to exclude those as well.

Oops, I wasn't able to find the break until CI finishes. I excluded those combinations as well: [1b19682](https://github.com/aws/aws-sdk-rails/pull/92/commits/1b19682abf096888be76700dc956749f9206e4e3)

As for the EOL versions, I respect your policy. 👍

ohbarye:
Thanks for your review too!

Issue #44, opened 2020-12-02
Using SESv2


OmriSama:
Hey guys! Only found out about this gem recently and I love it. Thanks for the effort!

I noticed the built in ActionMailer integration uses `SES` instead of the newer `SESv2` API. Is this going to make a huge difference? Are there plans to switch over?
Would you guys accept pull requests?

Thanks!

mullermp:
Hey @OmriSama , thanks for checking this gem out! We plan on making a blog post this week to highlight some of the new features we've added.

Regarding SES vs SESv2, I don't think there's an immediate difference in benefit between them in the context of using ActionMailer. SES hasn't officially announced any sort of deprecation AFAIK. At one point, a customer suggested Pinpoint but we held off because we weren't sure what the long term support plan was. Pinpoint would be interesting because it could support other channels other than email.

We would absolutely take a pull request on this and I can tag this as a feature request.

tmedford:
V2 of api support attachments greater than 10mb. Can we get the added now?

mullermp:
SESv2 support was added last month!

PR #93, opened 2022-12-06
SESv2 support

sgomez17:
*Issue #44 

*Description of changes:*
This is an updated version of PR #74, with the suggestions from that PR implemented.


lib/aws/rails/railtie.rb
```diff
@@ -28,11 +28,18 @@ class Railtie < ::Rails::Railtie
     #
     # @param [Symbol] name The name of the ActionMailer delivery method to
     #   register.
-    # @param [Hash] options The options you wish to pass on to the
-    #   Aws::SES::Client initialization method.
-    def self.add_action_mailer_delivery_method(name = :ses, options = {})
+    # @param [Hash] client_options The options you wish to pass on to the
+    #   Aws::SES[V2]::Client initialization method.
+    def self.add_action_mailer_delivery_method(name = :ses, client_options = {}, mailer = :ses)
```
mullermp, line 33: `name` already covers `mailer` here doesn't it? So I think we should re-use that one.

lib/aws/rails/sesv2_mailer.rb
```diff
@@ -0,0 +1,64 @@
...
+        send_opts = {}
+        send_opts[:content] = {}
+        send_opts[:content][:raw] = { data: message.to_s }
+
+        # Check for Rails >=6 compatibility
+        send_opts[:from_email_address] = if message.respond_to?(:from_address)
+                                           message.from_address&.to_s
+                                         else
+                                           Array.wrap(message.from).first&.to_s
+                                         end
```
mullermp, line 37: Do you have any documentation or information on this that I can look into? I am assuming this is Mail::Message

lib/aws/rails/sesv2_mailer.rb
```diff
@@ -0,0 +1,64 @@
...
+        send_opts[:destination][:bcc_addresses] = [*message.bcc]
+
+        send_opts[:configuration_set_name] = message.header['X-SES-CONFIGURATION-SET']&.yield_self do |field|
+          message.header.fields.delete(field).value
+        end
+
+        send_opts[:list_management_options] = message.header['X-SES-LIST-MANAGEMENT-OPTIONS']&.yield_self do |field|
+          contact_list_name, topic_name = message.header.fields.delete(field).value.sub("topic=", "").split(";").map(&:strip)
+          {contact_list_name: contact_list_name, topic_name: topic_name}.compact
+        end
```
mullermp, line 51: Can you explain what these are doing? And perhaps some of these should be placed in methods? `message.header.fields.delete(....)....` is very overloaded chaining behavior..

test/aws/rails/sesv2_mailer_test.rb
```diff
@@ -0,0 +1,80 @@
...
+        ActionMailer::Base.add_delivery_method(:sesv2, Sesv2Mailer, client_options)
+      end
+
+      describe '#settings' do
+        it 'returns an empty hash' do
+          expect(mailer.settings).to eq({})
+        end
+      end
+
+      describe '#deliver!' do
```
mullermp, line 62: I think we'll want to test the other fields and transformations you are doing - specifically from_email_address (Rails 6 vs older), list_management_options, and configuration_set_name. Let's ensure good test coverage of this class.

mullermp (dismissed):
Nice work! And thank you for taking ownership in this. This is a good start and improvement.

lib/aws/rails/sesv2_mailer.rb
sgomez17: Hi, yes they added some extensions for Mail::Message in rails, you can see them here:
https://github.com/rails/rails/blob/ade27e5028ccbaaa5db8d966500844017925a751/actionmailbox/lib/action_mailbox/mail_ext/addresses.rb

I can only see those extensions from rails v6 onwards, and I noticed it just because the rails app I'm maintaining is still on v5.2.7.1, and threw an error when trying to read "message.from_address". I could add the extensions just to our app, but I thought that simple check might make this gem backwards compatible with Rails 5.2.

sgomez17:
hi @mullermp, actually just noticed your first comment in #74. Would you suggest replacing the aws-sdk-ses with aws-sdk-sesv2 completely? By the look of it, we only use .send_raw_email from the SES SDK, which can be replaced with the .send_email method in SESV2 setting the content as :raw, which is what this PR does. Let me know and I can update this PR to completely replace SES with SESV2, instead of supporting both in parallel.

lib/aws/rails/railtie.rb
mullermp: Ah sorry.. brain fart. This one is weird. Previously someone can have name :foo for SES mailer V1. We could decide if "name == :sesv2" means also using ses v2 mailer, and tie delivery method name to mailer class. What do you think?

lib/aws/rails/sesv2_mailer.rb
mullermp: Thanks. What about `message.from` and why doesn't this work generically for any Rails version, even newer ones?

lib/aws/rails/railtie.rb
sgomez17: Yes I think that 'mailer' parameter is redundant to be honest. I'll amend the code to use "name == :sesv2" as the indicator to use V2.

lib/aws/rails/sesv2_mailer.rb
sgomez17: I think that `message.from` should be ok to use from looking at the comments on the code for [mail/message](https://github.com/mikel/mail/blob/bf02b6772e4df20e165c008f595f4eadb8c595a7/lib/mail/message.rb#L658).
But I am not very familiar with the `mail` gem, so please take a look and let me know any thoughts. 
I'm happy to implement the changes an update this PR. My suggestion is to go with `message.from`. I just put that check as the other PR #74 used `message.from_address`, but actually I think we can replace that with `message.from`.

sgomez17:
Yes ok, I think it's fine to support both versions in parallel for now, to ensure backwards compatibility. But I will drop the :mailer parameter, and just use 'name == :sesv2' to detect use of V2.

lib/aws/rails/sesv2_mailer.rb
mullermp: Yeah, I think message.from might work generically and we won't need to branch here. That makes sense that you carried it over originally. Can you confirm with your testing (assuming you are doing this locally)?

lib/aws/rails/railtie.rb
mullermp: Sounds good, see my comment below about checking for this.

mullermp:
This PR https://github.com/aws/aws-sdk-rails/pull/94 adds SMTP support to V1 mailer - it may be useful to apply to V2's

lib/aws/rails/sesv2_mailer.rb
```diff
@@ -0,0 +1,64 @@
...
+        send_opts[:from_email_address] = if message.respond_to?(:from_address)
+                                           message.from_address&.to_s
+                                         else
+                                           Array.wrap(message.from).first&.to_s
+                                         end
+
+        send_opts[:destination] = {}
+        send_opts[:destination][:to_addresses] = [*message.to]
+        send_opts[:destination][:cc_addresses] = [*message.cc]
+        send_opts[:destination][:bcc_addresses] = [*message.bcc]
```
c960657, line 42: `Mail::Message#to`, `#cc` and `#bcc` reflects the `To`, `Cc` and `Bcc` mail header fields, but the mail should not necessarily be sent to the addresses specified in the header. You should use `Mail::Message#smtp_envelope_to` instead. See #94.

sgomez17: Thanks @c960657. Any suggestions for this? According to the [AWS docs](https://docs.aws.amazon.com/ses/latest/APIReference-V2/API_Destination.html), we need to provide To/Cc/Bcc separately, but the smtp_envelope_to is a combination of all those fields right? So I'm not entirely sure how to use smpt_envelope_to here.

image

c960657: It is unclear from the docs what these fields are used for, so it requires some experimenting. I would try just sticking all `smtp_envelope_to` addresses in the Bcc field and leave the others empty.

mullermp: I'll need to double check both APIs too. Thanks for bringing this up.

mullermp:
I get the same behavior for SESv2 - when using smtp_envelope_to/from as `from_email_address` and `destination[:to_addresses]` in the API:
```
sesv2.send_email(from_email_address: m.smtp_envelope_from, destination: { to_addresses: m.smtp_envelope_to }, content: { raw: { data: m.to_s } })
```
The email is from and delivered by the To/From headers. In SESv2 the API seems to be multi-mode, allowing for raw, template, and simple sending. Unfortunately raw message is required for attachments and other useful features - and the [documentation](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SESV2/Client.html#send_email-instance_method) seems to indicate that it only cares about the body in the raw case.

mullermp:
I'm trying to reach out to someone on SES to get more information about this.

sgomez17:
Thanks @mullermp, in the interim I will update the PR with tests for those other headers (config-sets and list-management-options).

sgomez17:
@mullermp I have added some tests. Still pending what to do about using .smtp_envelope_to.
Also I am unable to reproduce those tests that keep failing in the CI build. Locally all tests pass, even with the given ruby versions that the build fails for. Could you give me a hand with that or point me in the right direction?

mullermp:
Locally it looks like tests will pass or fail if you run them multiple times (indeterministic). I'll try to help track that down.

mullermp:
Can you grant me permission to push to this PR and fork? `ERROR: Permission to sgomez17/aws-sdk-rails.git denied to mullermp.`

mullermp:
As far as getting tests to go, I have a few ideas. I think we need to add `Aws::Rails.add_action_mailer_delivery_method(:sesv2)` to Railtie initializer so that it's available generally. The `before` blocks in both mailer tests may need to be refactored to use `Aws::Rails.add_action_mailer_delivery_method(:ses, client_options)` - however the way we deliver and expect data in the tests may need to change (or how those pieces are hooked up)

sgomez17:
Sorry @mullermp I missed your message last week, but I have ticked the "Allow edits by maintainers box". Let me know if you have any issues pushing to this PR.

mullermp:
I think there's order dependence in tests, somewhere. Running with a specific seed I can reliably reproduce some errors (it's a region error - some seeds give invalid delivery method). `SEED=45610 BUNDLE_GEMFILE=gemfiles/rails-6.0.gemfile bundle exec rake`

mullermp:
This was pretty silly - TestMailer was redefined in the v2 mailer tests - so ordering was dependent upon that class being defined (and re-defined). Now failures are usage of `yield_self` in older Ruby.

mullermp:
I think that since we are sending a raw email with `message.to_s`, we don't need to set `from_email_address`, `destination`, `configuration_set_name`, and `list_management_options`, because the raw email feature will take that directly from headers. From my earlier testing of the SMTP envelope to/from, those parameters didn't seem to matter for raw email. Do you have use cases with these headers and could confirm that?

sgomez17:
No @mullermp I don't have a use case for those headers, that was part of the original PR (#74) that's why I kept them. We could remove them from now, and if someone down the track needs them they can raise a separate issue/PR. Thoughts?

sgomez17:
And regarding .yield_self issues, again it's from the original PR, but I can easily refactor that to make it backwards compatible, although won't need to if we decide to drop support for those headers.

mullermp:
@sgomez17 @c960657 

I've refactored both mailers to something that seems minimal and correct. Through my testing, it seems that SES unfortunately ignores members like sources, destinations, configuration sets, etc, when using "raw" email mode - which is necessary for attachment support. I'm still trying to get SES to look at my internal ticket to get more information about this. I definitely agree that SMTP envelope address should be configurable and possibly different than the message to/from. I will push on the SES team. For now I think omitting everything except the raw message in the API call is correct, and leaves a pretty simple implementation. In testing, SES did validate the names of my configuration set and contact lists, using the x-amz custom headers in the message, so I believe that to be correct and we simply want to check for pass through. Please also double check this .. I definitely don't want to break existing customers such as yourself!

sgomez17:
Hi @mullermp, I found an issue but not sure whether it is related to the latest changes. Bcc doesn't seem to be working when using either :ses or :sesv2 delivery methods:
I am sending an email, ensuring that :bcc is set, and both delivery methods seem to be ignoring it. I get the email in the CC mailbox, but not BCC. I have tried swapping them to ensure it's not just on my email provider side, and CC works reliably but BCC doesn't work at all. I'm going to try to debug it, but you might know straight away what could be wrong.
image

mullermp:
How are you creating the mail object? (Where/how are you passing bcc?)

sgomez17:
I was just hard-coding it in one of our mailers like this:
image

But now even in the tests that we have for this PR, look at this:
image

mullermp:
That makes sense. I'll test and update it tomorrow.

mullermp (approved):
I think it makes sense now. I did not know bcc was not in the header (It makes total sense, otherwise it's not blind .. ) I was able to get bcc working when we were somewhat "complete" with providing params. Thank you for double checking.

sgomez17:
Thanks @mullermp, I can confirm that bcc is working as expected with the latest commit. I haven't found any other issues for our use case.

alextwoods (approved):
Change LGTM.  

My only suggestion is to update the README.md Mailer section.

mullermp:
We'll release this after the new year. Thank you for your contribution!

PR #94, opened 2022-12-07
Support `Mail::Message#smtp_envelope_from` and `_to`

c960657:
SMTP allows sending mail using different sender and recipient addresses than those present in the `From`/`To`/`Cc`/etc. message header fields.

Use `Mail::Message#smtp_envelope_from` and `#smtp_envelope_to` to retrieve the `Source` and `Destination` parameters for `SendRawEmail`. If not specified explicitly, these addresses default to the values of the `Return-Path`/`Sender`/`From` and `To`/`Cc`/`Bcc` headers.

These methods were present in mail 2.5, the [oldest version supported by Rails 5.2](https://github.com/rails/rails/blob/5-2-stable/actionmailer/actionmailer.gemspec#L33), which is the oldest version supported by aws-sdk-rails.

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

### DIFF
--- a/test/aws/rails/mailer_test.rb
+++ b/test/aws/rails/mailer_test.rb
@@ -7,13 +7,9 @@ class TestMailer < ActionMailer::Base
   layout nil
 
   def deliverable(options = {})
-    mail(
-      body: options[:body],
-      delivery_method: :ses,
-      from: options[:from],
-      subject: options[:subject],
-      to: options[:to]
-    )
+    mail(delivery_method: :ses, body: '').tap do |m|
+      options.each {|k, v| m.send(k, v) }
+    end
   end
 end
 
@@ -36,8 +32,10 @@ module Aws
         TestMailer.deliverable(
           body: 'Hallo',
           from: 'sender@example.com',
+          smtp_envelope_from: 'envelope-sender@example.com',
           subject: 'This is a test',
-          to: 'recipient@example.com'
+          to: 'recipient@example.com',
+          smtp_envelope_to: 'envelope-recipient@example.com',
         )
       end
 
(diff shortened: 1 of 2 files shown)


test/aws/rails/mailer_test.rb
```diff
@@ -7,13 +7,9 @@ class TestMailer < ActionMailer::Base
...
-    mail(
-      body: options[:body],
-      delivery_method: :ses,
-      from: options[:from],
-      subject: options[:subject],
-      to: options[:to]
-    )
+    mail(delivery_method: :ses, body: '').tap do |m|
+      options.each {|k, v| m.send(k, v) }
+    end
```
mullermp, line 12: Making this generic is fine. Could we pass those directly to mail's parameter instead of using send on accessor methods?

```
mail(**{delivery_method: :ses, body: ''}.merge(options))
```

mullermp (commented):
Thanks for opening this PR! This change overall looks good to me. I think we would want this on other mailers too, like SES v2, which is [currently in PR](https://github.com/aws/aws-sdk-rails/pull/93) if you are interested in contributing.

test/aws/rails/mailer_test.rb
```diff
@@ -36,8 +32,10 @@ module Rails
         TestMailer.deliverable(
           body: 'Hallo',
           from: 'sender@example.com',
```
mullermp, line 34: Do you think we should add a test that checks the defaults of from/to methods when not using smtp methods?

test/aws/rails/mailer_test.rb
c960657: I tried that, but that only works for `body` and attributes that are mapped to header fields, but not for `smtp_envelope_from`/`_to`. But now I have made a separate test for SMTP envelope addresses, so this is limitation is no longer a problem.

mullermp:
Thanks for the contribution! I will probably try and release this next week if that's ok. Hoping to get some other changes in (SES v2)

Issue #97, opened 2023-01-19
Sender name gets dropped from email_from when using SESv2


sgomez17:
Following the introduction of SESV2, we have noticed that the name part of the sender gets dropped when using SESV2 when then email is sent.
So if the email comes from `Some Name <some.name@test.com>`, the recipient only sees the email address 'some.name@test.com>', but the 'Some Name' part is dropped.
We tested the same email with SES (v1), and that works ok and the email is received with both parts of the sender email (name + address).

sgomez17:
Upon further investigation I can see that none of the :from methods from Mail::Message expose the name part of the sender, only the address.
![image](https://user-images.githubusercontent.com/33270615/213367863-bf8e1c8d-5e38-4051-ae5a-c87be1a54a2a.png)

However the header does contain the name.
![image](https://user-images.githubusercontent.com/33270615/213368423-41bf9cf3-88cf-46ba-822e-f835fb1e6498.png)

Maybe this is why in rails they extended the Mail::Message class from Rails6, as shown here: https://github.com/rails/rails/blob/ade27e5028ccbaaa5db8d966500844017925a751/actionmailbox/lib/action_mailbox/mail_ext/addresses.rb

We could introduce a check for mailer_sesv2.rb, if message responds to :from_address then use it, otherwise use the current value, as suggested in the first part of #93. Thoughts? I'm happy to create a PR for this.

mullermp:
I think that makes sense. If you start a PR, I can help shepherd it out. I'll be on leave for a few months very soon - we can try to get this fixed before I go.

mullermp:
Just to clarify, does this work for SESv1 *in the most recent aws-sdk-rails version*? Both implementations appear to be the same.

sgomez17:
Yes, in the most recent version of this gem, I tried with :ses and :sesv2. Here are the message details for both emails, you can see the SESv2 version is missing the sender name:
![image](https://user-images.githubusercontent.com/33270615/213569758-da6b76f2-76fc-4aab-b0df-5870dc3c4765.png)

alextwoods:
The PR has been merged and the fix has been released as part of 3.7.0.

PR #98, opened 2023-01-20
Fix sesv2 from_email_address missing sender name

sgomez17:
*Issue #97*

*Description of changes:*
Fix the issue reported in #97 for SESV2.

### DIFF
--- a/lib/aws/rails/sesv2_mailer.rb
+++ b/lib/aws/rails/sesv2_mailer.rb
@@ -27,7 +27,7 @@ module Aws
       def deliver!(message)
         params = {
           content: { raw: { data: message.to_s } },
-          from_email_address: message.smtp_envelope_from, # defaults to From header
+          from_email_address: from_email_address(message),
           # defaults to destinations (To,Cc,Bcc) - stick all on bcc to be delivered anyway
           destination: { bcc_addresses: message.smtp_envelope_to }
         }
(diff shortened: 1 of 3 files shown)


lib/aws/rails/sesv2_mailer.rb
```diff
@@ -27,7 +27,7 @@ def initialize(options = {})
       def deliver!(message)
         params = {
           content: { raw: { data: message.to_s } },
-          from_email_address: message.smtp_envelope_from, # defaults to From header
+          from_email_address: from_email_address(message),
           # defaults to destinations (To,Cc,Bcc) - stick all on bcc to be delivered anyway
           destination: { bcc_addresses: message.smtp_envelope_to }
```
mullermp, line 32: Do we need a similar fix to smtp_envelope_to somehow?

mullermp:
I'm doing some testing today. I'm not sure about this approach currently. The correct sender name appears when I set the From header with a name and email address AND omit `smtp_envelope_from`. We may actually want to check `mail.from_addrs.first != mail.smtp_envelope_from`, and when true, use `mail.smtp_envelope_from` as ses v2 `from_email_address` param, or omit it otherwise.

mullermp:
I think I'm leaning towards a solution that uses `instance_variable_get` on smtp_envelope_to/from values and omitting / adapting certain SES v2 params.

mullermp:
Is the SESv2 mailer working with Bcc? I thought we solved that problem already, but in all of my testing, and even just using what's coded now, it's not working... SES v2 is so finicky :/

mullermp:
@sgomez17 I pushed a change with the `instance_variable_get` approach. I think the defaults for `smtp_envelope_from/to` methods are not ideal for SES v2's API design, but work well for SES v1. I think in both APIs, SES will do comparisons and defaults to find the correct names and addresses. We can check if the instance variable was set (they were explicitly set) and if they were, use them as `from_email_address` and `destinations[:to_addresses]`. Otherwise, we omit `from_email_address` (SES uses From header by default, including name) and we use To header value. I think this is correct because SES v2 documents all of these members as optional - because raw message is supposed to define all of these ourselves.

Please try this mailer out and see if it works for you! I did some testing on my side and it seems to work as expected - I see names for To, From, and Cc, and also sends to Bcc.

sgomez17:
hi @mullermp, I have tested it and yes it and that latest version works well for our use-case, so it's a thumbs up from me to merge :+1:

PR #102, opened 2023-09-26
Add support for selectively choosing deduplication keys.

Kytozz:
*Description of changes:*

Previously, when using ContentBasedDeduplication, only job_id was excluded from the deduplication key. However, depending on the requirements of Rails applications, it may be necessary to exclude keys other than job_id as well.

I have addressed this issue by allowing the deduplication key to be customized in the configuration or within each job class. 
Please note that, by default, only job_id is excluded.


lib/active_job/queue_adapters/amazon_sqs_adapter.rb
```diff
@@ -54,6 +51,25 @@ def message_attributes(job)
           }
         }
       end
+
+      def deduplication_body(job, body)
+        dedup_keys = job.deduplication_keys if job.respond_to?(:deduplication_keys)
+        dedup_keys ||= Aws::Rails::SqsActiveJob.config.deduplication_keys.map(&:to_s)
```
mullermp, line 57: I think it makes sense to use symbol throughout the method (including for job_id), and then convert to string when doing body.slice. It would eliminate to_s here and in deduplicate_with.

lib/aws/rails/sqs_active_job/configuration.rb
```diff
@@ -81,13 +83,18 @@ class Configuration
         #   +active_job.queue_adapter = :amazon_sqs_async+.  Called with:
         #   [error, job, job_options]
         #
-        # @option options [SQS::Client] :client SQS Client to use.  A default
+        # @option options [SQS::Client] :client SQS Client to use. A default
         #   client will be created if none is provided.
+        #
+        # @option options [Array] :deduplication_keys ([])
```
mullermp, line 89: It'd be nice to convey the default here (all properties except job id) since it is not actually an empty array, or maybe the configuration should be changed to be an exclude list instead of an override (which job id can be hard coded appended to), whichever has better customer experience.

lib/aws/rails/sqs_active_job/deduplication.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Aws
+  module Rails
+    module SqsActiveJob
+      extend ActiveSupport::Concern
```
mullermp, line 6: Good use of concern here. Are there any other SQS active job related examples in this gem we can clean up as such?

test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
```diff
@@ -29,20 +29,37 @@ module QueueAdapters
...
-                                  message_deduplication_id: instance_of(String)
-                                }
-                              )
+          expect(client).to receive(:send_message).with(
+            {
+              queue_url: 'https://queue-url.fifo',
+              message_body: instance_of(String),
+              message_attributes: instance_of(Hash),
+              message_group_id: Aws::Rails::SqsActiveJob.config.message_group_id,
+              message_deduplication_id: instance_of(String)
```
mullermp, line 38: Can we try to assert the default and the override deduplication_id below to ensure that the configuration is working as expected?

mullermp (commented):
Thanks for opening a pull request. I think this is a valid feature to consider.

mullermp:
It looks like test failures are related to MiniTest somehow. The JRuby 9.4 failures appear to be either some deprecation or bug.

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
```diff
@@ -54,6 +51,25 @@ def message_attributes(job)
...
+
+        ignored_dedup_key = 'job_id'
+
+        if dedup_keys.include?(ignored_dedup_key)
```
alextwoods, line 61: It might be better to check for an invalid key upfront, when its set rather then every time its used (raise the error where its being made) - I think you could do that in the setter for deduplication_keys.

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
```diff
@@ -27,11 +27,8 @@ def _enqueue(job, body = nil, send_message_opts = {})
         send_message_opts[:message_attributes] = message_attributes(job)
 
         if Aws::Rails::SqsActiveJob.fifo?(queue_url)
-          # job_id is unique per initialization of job
-          # Remove it from message dup id to ensure run-once behavior
-          # with ActiveJob retries
           send_message_opts[:message_deduplication_id] =
-            Digest::SHA256.hexdigest(Aws::Json.dump(body.except('job_id')))
+            Digest::SHA256.hexdigest(Aws::Json.dump(deduplication_body(job, body)))
```
alextwoods, line 31: A general thought - rather than configuring the keys to either include or exclude - would it make sense to allow the user to configure a lambda/proc that would be called? That would allow full customization of the deduplication_id.  (But may be more complex for most use cases....)

lib/aws/rails/sqs_active_job/configuration.rb
alextwoods: I'd personally lean towards an exclude list instead of include list.  That way the default doesn't need to change if other fields are added in the future.  And in general, I'd imagine most use cases would want to just exclude 1 or 2 additional fields?

alextwoods (commented):
Overall looks good and I think is a useful feature - thanks for submitting!

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
mullermp: This is possibly overkill ..

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
Kytozz: Use to_s within each definition to uniquely add job_id to keys within the deduplicate_without method and excluded_deduplication_keys getter definitions.
Anyway, I agree with your ideas and would be happy to hear any other good ideas.

lib/aws/rails/sqs_active_job/configuration.rb
Kytozz: Revised with feedback from alextwoods :)

test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
Kytozz: I added some specs.

lib/aws/rails/sqs_active_job/configuration.rb
```diff
@@ -81,16 +84,25 @@ class Configuration
         #   +active_job.queue_adapter = :amazon_sqs_async+.  Called with:
         #   [error, job, job_options]
         #
-        # @option options [SQS::Client] :client SQS Client to use.  A default
+        # @option options [SQS::Client] :client SQS Client to use. A default
         #   client will be created if none is provided.
+        #
+        # @option options [Array] :excluded_deduplication_keys ('job_id')
```
mullermp, line 90: Default should be `(['job_id'])` since it's Array

test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
```diff
@@ -29,20 +29,71 @@ module QueueAdapters
...
+            end
+            it 'adds customized message_deduplication_id' do
+              expect_any_instance_of(TestJobWithDedupKeys).to receive(:excluded_deduplication_keys).once
+              expect(client).to receive(:send_message).with(
+                {
+                  queue_url: 'https://queue-url.fifo',
+                  message_body: instance_of(String),
+                  message_attributes: instance_of(Hash),
+                  message_group_id: Aws::Rails::SqsActiveJob.config.message_group_id,
+                  message_deduplication_id: instance_of(String)
```
mullermp, line 62: Is there a way to strongly assert that message_deduplication_id does not contain job_class and queue_name here? Maybe an implementation detail but we could assert `Digest::SHA256.hexdigest` or `Aws::Json.dump` doesn't receive them.

mullermp (approved):
Nice! Just some minor comments and I'll take a look at fixing CI ..

mullermp:
The test failures in main should be fixed.

mullermp:
I pushed a change to main that configured and ran rubocop on everything, as well as bump the minimum rails version to 2.3. Please check to see if main merged in correctly and didn't stomp on any of your changes.

mullermp:
Please also fix your rubocop errors (new CI I've added)

Kytozz:
Confirmation about Rubocop.

```
> bundle e rubocop -E -S 
===>
test/active_job/queue_adapters/amazon_sqs_adapter_test.rb:7:3: C: Metrics/ModuleLength: Module has too many lines. [144/112]
  module QueueAdapters ...
  ^^^^^^^^^^^^^^^^^^^^
test/active_job/queue_adapters/amazon_sqs_adapter_test.rb:8:5: C: Metrics/BlockLength: Block has too many lines. [142/110]
    describe AmazonSqsAdapter do ...
```

Rubocop also limits the number of lines in the test files, but this may be an unnecessary calibration due to the functionality of RSpec. RSpec is written in a dedicated DSL, which inevitably increases the number of lines per block.

if you agree with this, I will add the following code to .rubocop.yml:

```yml
# .rubocop.yml

Metrics/BlockLength:
  Exclude:
    - 'test/**/*.rb'

Metrics/ModuleLength:
  Exclude:
    - 'test/**/*.rb'
```

Kytozz:
Note that Rubocop CI still detects errors unless https://github.com/aws/aws-sdk-rails/pull/102#issuecomment-1738320788 is applied or fixed individually.

mullermp:
Yeah, it's fine to add to to-do or global ignore.

Kytozz:
https://github.com/aws/aws-sdk-rails/pull/102#issuecomment-1738320788
Added it to global ignore :)

mullermp (approved):

mullermp:
Thanks for your contribution! I will get this released tomorrow.

mullermp:
Actually, I was able to do this today, please check to see if it works for you!

kyto64:
Thanks!

PR #109, opened 2024-01-17
supports ruby3.3

kakubin:


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Officially supports Ruby version 3.3.
```
mullermp, line 4: This changelog entry isn't necessary - It's already supported since we have minimum Ruby versions of the gem, but it's useful to have it in CI.

mullermp (approved):

Issue #105, opened 2023-10-10
Support for `ActiveJob.perform_all_later` (Rails 7.1)


gondalez:
Hi, 

Thanks for the well written gem. We have been using it for ActiveJob mainly and it has been pleasantly performant and unsurprising with our production workload 😌 🙏 

I stumbled on `ActiveJob.perform_all_later` that was [added in rails 7.1](https://www.bigbinary.com/blog/rails-7-1-adds-activejob-perform-all-later).

Are there any plans to support it in `aws-sdk-rails`?

Specifically to support creating many sqs message in a single [`SendMessageBatch`](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html) for increased throughput.

The rails PR for reference: https://github.com/rails/rails/pull/46603

mullermp:
Thanks for opening an issue. Adding this as a feature request.

PR #110, opened 2024-01-17
Support AJ 7.1 perform_all_later api

kakubin:
[Rails 7.1 adds API for queuing multiple jobs simultaneously.](https://github.com/rails/rails/pull/46603)
To use this feature, we need to implement `#enqueue_all` method in job adapter.

references:
- https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html#API_SendMessageBatch_RequestSyntax
- https://github.com/aws/aws-sdk-ruby/blob/1544a183c0e78cf22fe1889805644d00880bfc4e/gems/aws-sdk-sqs/lib/aws-sdk-sqs/client.rb#L2277-L2376

close: #105

### DIFF
--- a/test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
+++ b/test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
@@ -188,6 +188,38 @@ module ActiveJob
           end.to raise_error ArgumentError
         end
       end
+
+      if Gem::Version.new(Rails::VERSION::STRING) >= Gem::Version.new('7.1.0')
+        describe 'with multiple jobs' do
+          before do
+            response = double('Response')
+            allow(response).to receive(:successful).and_return([1, 2])
+            allow(client).to receive(:send_message_batch).and_return(response)
+          end
+
+          it do
+            expect(client).to receive(:send_message_batch).with(
+              {
+                queue_url: 'https://queue-url',
+                entries: [
+                  {
+                    id: instance_of(String),
+                    message_body: instance_of(String),
+                    message_attributes: instance_of(Hash)
+                  },
+                  {
+                    id: instance_of(String),
+                    message_body: instance_of(String),
+                    message_attributes: instance_of(Hash)
+                  }
+                ]
+              }
+            ).once
+
+            ActiveJob.perform_all_later(TestJob.new('test'), TestJob.new('test'))
+          end
+        end
+      end
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Support `enqueue_all` in the SQS ActiveJob adapter. 
+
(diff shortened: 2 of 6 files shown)


test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
```diff
@@ -188,6 +188,30 @@ module QueueAdapters
           end.to raise_error ArgumentError
         end
       end
+
+      if Gem::Version.new(Rails::VERSION::STRING) >= Gem::Version.new('7.1.0')
```
mullermp, line 192: One option we can consider is major version bumping aws-sdk-rails to hard depend on Rails 7.1, and dropping old dependencies. I would not be against this.

Gemfile
```diff
@@ -12,6 +12,7 @@ group :test do
   gem 'bcrypt'
   gem 'rspec-expectations'
   gem 'rspec-mocks'
+  gem 'timecop'
```
mullermp, line 15: I would prefer if we didn't have to add extra test dependencies. The way this is used in specs seems simple enough that we could just mock the expectations.

mullermp (commented):
Thanks for opening a pull request. This is a great first start!

test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
kakubin: This change has no effect on the old rails version, so it could be tagged as minor change. 
but if there is no other opportunity like this PR(supports Rails 7.1 feature), I think it would be a good idea to cut off old version's support and major bumping with this change.

mullermp: I'll talk with the team and get back to you.

mullermp:
I think we are leaning towards a major version bump and dropping deprecated rails. I'll take a deeper look this and next week.

Regardless of this decision, would you be interested in adding a real "integration test" for this feature? This repo should have a sample rails app that minimally sets up each basic feature for testing. It would be nice to have a basic working example and document it too.

test/active_job/queue_adapters/amazon_sqs_adapter_test.rb
mullermp: I think a minor version release is fine for this. Checking the dependencies, they're already loose.

mullermp (approved):
Thanks for your contribution!

alextwoods (approved):
Looks good - Thanks for the contribution!

kakubin:
@mullermp 
> Would you be interested in adding a real "integration test" for this feature? This repo should have a sample rails app that minimally sets up each basic feature for testing. It would be nice to have a basic working example and document it too.

I'll do 👍 
Are you saying adding specs after this PR merged?
Also, if I add spec for this feature, we need to upgrade Rails version on `sample_app/`.

mullermp:
Yes - if you'd like to help upgrade the sample app and add a small job that demonstrates the feature, that would be very helpful. The sample app is very minimal so I don't see issue with upgrading it.

PR #115, opened 2024-02-28
Add retry_standard_errors config for SQS ActiveJob

### LINKED ISSUE #114: SQS ActiveJobs are retried if uncaught exceptions are raised or retry attempts are exceeded
https://guides.rubyonrails.org/active_job_basics.html#retrying-or-discarding-failed-jobs notes that failed jobs are not retried unless the jobs are configured otherwise, however SQS backed ActiveJobs that raise exceptions which are not explicitly discarded or retried continue to be run. This occurs with either the `amazon_sqs` or `amazon_sqs_async` adapters. For example, a simple job such as:

```ruby
class SampleJob < ActiveJob::Base
  queue_as :default

  def perform
    raise Exception, "testing"
  end
end
```

never gets deleted from the queue, and will be fetched and run again after the messages visibility_timeout expires. I am not sure if this is desired behavior or not, but it caught me off guard as it is different than the above note in the ActiveJob rails docs. A similar test with the Resque gem (being the only other queueing service I have familiarity with and an active project using) removed the job after a single failed run.

If this is intended behavior of the Rails SQS ActiveJob backend, a note in the Readme that the behavior differs from the rails noted behavior would be helpful.

**Versions:**
Rails 7.1.3
Ruby 3.2.2
aws-sdk-rails 3.10.0
SQS standard queue (not FIFO)
OSX Ventura 13.6.4

*edit:*
It looks like even if a `retry_on Exception, wait: 40.seconds, attempts: 2` is added to the job, it's not removed after 2 attempts as would be expected. Are retries with limited attempts not supported with the SQS backend?

alextwoods:
*Issue #, if available:* 
Fixes #114 

*Description of changes:*
Adds configuration to change the behavior of SQS ActiveJob's handling of retries for exceptions from ActiveJobs.  

The existing behavior does not follow the [https://guides.rubyonrails.org/active_job_basics.html#retrying-or-discarding-failed-jobs](Rails ActiveJob) retry/discard behavior which notes that failed jobs are not retried unless the jobs are configured otherwise.

To avoid this being a breaking change for existing users, the flag currently defaults to `true` (preserving the existing behavior). 

Pending review, I will update the Readme to describe in more detail the interaction of SQS retry and Rails ActiveJob retries.

### DIFF
--- a/lib/aws/rails/sqs_active_job/configuration.rb
+++ b/lib/aws/rails/sqs_active_job/configuration.rb
@@ -25,7 +25,7 @@ module Aws
         DEFAULTS = {
           max_messages: 10,
           shutdown_timeout: 15,
-          retry_standard_errors: true,
+          retry_standard_errors: true, # TODO: Remove in next MV
           queues: {},
           logger: ::Rails.logger,
           message_group_id: 'SqsActiveJobGroup',
--- a/lib/aws/rails/sqs_active_job/configuration.rb
+++ b/lib/aws/rails/sqs_active_job/configuration.rb
@@ -25,6 +25,7 @@ module Aws
         DEFAULTS = {
           max_messages: 10,
           shutdown_timeout: 15,
+          retry_standard_errors: true,
           queues: {},
           logger: ::Rails.logger,
           message_group_id: 'SqsActiveJobGroup',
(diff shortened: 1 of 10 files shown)


lib/aws/rails/sqs_active_job/configuration.rb
```diff
@@ -25,6 +25,7 @@ class Configuration
         DEFAULTS = {
           max_messages: 10,
           shutdown_timeout: 15,
+          retry_standard_errors: true,
```
mullermp, line 28: Do we really need to preserve backwards compatibility for this? It seems safe to not rely on the fact that something failed and retried implicitly.

mullermp (approved):

lib/aws/rails/sqs_active_job/configuration.rb
alextwoods: I added retry_standard_errors with a default of true because right now, users are getting a built in retry of everything and they could be relying on that to retry expected, transient errors.  If they set it to false (or if we default it to false), any of those errors that were previously retried will now fail immediately on the first error.

I've also added a check on `execution_exceptions` and we will NOT retry if ActiveJob has already caught/handled exceptions (note, this is true both for `retry_on` and `discard_on`).  I think this is a reasonable compromise that address the original issue while not removing existing retry behavior that users may be relying on.

mullermp (approved):

PR #116, opened 2024-03-29
Add docker with cgroup2 detection

bongole:
*Issue #, if available:*
* Amazon Linux 2023 runs docker in cgroup2, so EbsSqsActiveJobMiddleware#app_runs_in_docker_container? which only assumes cgroup1 will return false.

*Description of changes:*
1. Refactored the code of detecting cgroup1 using external commands and add its tests.
2. Add the code  of detecting cgroup2 and add its tests.

### DIFF
--- a/lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
+++ b/lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
@@ -85,14 +85,14 @@ module Aws
       end
 
       def app_runs_in_docker_container?
-        @app_runs_in_docker_container ||= in_docker_containr_with_cgroup1? || in_docker_containr_with_cgroup2?
+        @app_runs_in_docker_container ||= in_docker_container_with_cgroup1? || in_docker_container_with_cgroup2?
       end
 
-      def in_docker_containr_with_cgroup1?
+      def in_docker_container_with_cgroup1?
         File.exist?('/proc/1/cgroup') && File.read('/proc/1/cgroup') =~ %r{/docker/}
       end
 
-      def in_docker_containr_with_cgroup2?
+      def in_docker_container_with_cgroup2?
         File.exist?('/proc/self/mountinfo') && File.read('/proc/self/mountinfo') =~ %r{/docker/containers/}
       end
 
--- a/lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
+++ b/lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
@@ -93,7 +93,7 @@ module Aws
       end
 
       def in_docker_containr_with_cgroup2?
-        File.exist?('/proc/1/mountinfo') && File.read('/proc/1/mountinfo') =~ %r{/docker/containers/}
+        File.exist?('/proc/self/mountinfo') && File.read('/proc/self/mountinfo') =~ %r{/docker/containers/}
       end
 
       def default_gw_ips
--- a/lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
+++ b/lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
@@ -85,7 +85,15 @@ module Aws
       end
 
       def app_runs_in_docker_container?
-        @app_runs_in_docker_container ||= `[ -f /proc/1/cgroup ] && cat /proc/1/cgroup` =~ /docker/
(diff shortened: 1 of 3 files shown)


lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
```diff
@@ -85,7 +85,15 @@ def sent_from_docker_host?(request)
...
-        @app_runs_in_docker_container ||= `[ -f /proc/1/cgroup ] && cat /proc/1/cgroup` =~ /docker/
+        @app_runs_in_docker_container ||= in_docker_containr_with_cgroup1? || in_docker_containr_with_cgroup2?
+      end
+
+      def in_docker_containr_with_cgroup1?
+        File.exist?('/proc/1/cgroup') && File.read('/proc/1/cgroup') =~ %r{/docker/}
+      end
+
+      def in_docker_containr_with_cgroup2?
+        File.exist?('/proc/1/mountinfo') && File.read('/proc/1/mountinfo') =~ %r{/docker/containers/}
```
mullermp, line 96: This [article](https://www.baeldung.com/linux/is-process-running-inside-container) suggests `/proc/self/mountinfo`

lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
```diff
@@ -85,7 +85,15 @@ def sent_from_docker_host?(request)
       end
 
       def app_runs_in_docker_container?
-        @app_runs_in_docker_container ||= `[ -f /proc/1/cgroup ] && cat /proc/1/cgroup` =~ /docker/
+        @app_runs_in_docker_container ||= in_docker_containr_with_cgroup1? || in_docker_containr_with_cgroup2?
```
mullermp, line 88: Is there a better way other than cgroups (i.e. `dockerenv` file)? Based on my searching, it seems to be the best, but all options don't seem great. 

I'm open to re-evaluating this.

mullermp (commented):
I think the change makes sense.

lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
bongole: Thanks for reviewing.
Fixed in the following commit.
https://github.com/aws/aws-sdk-rails/pull/116/commits/845580bcfc4a092d28ba5a2f245112c272d151d9

lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
bongole: I have looked for a way to detect if a process is running inside a docker container, 
but there doesn't seem to be an official and portable way to detect it.

Maybe reading /proc/1/cgrops and /proc/self/mountinfo is the best workaround now.

References:
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/6694
* https://github.com/containerd/containerd/issues/8185
* https://github.com/opencontainers/runtime-spec/issues/1105
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/CgroupV2ContainerIdExtractor.java
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/resources/library/src/main/java/io/opentelemetry/instrumentation/resources/CgroupV1ContainerIdExtractor.java

lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
```diff
@@ -85,7 +85,15 @@ def sent_from_docker_host?(request)
...
+      end
+
+      def in_docker_containr_with_cgroup1?
```
mullermp, line 91: Not sure if you noticed, but this should be `container` (you forgot the `e`)

alextwoods (approved):
LGTM!  Thanks for the contribution!

mullermp (approved):

lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
bongole: oh, it's typo.
thanks.
Fixed in the following commit.
https://github.com/aws/aws-sdk-rails/pull/116/commits/21c62ec8e9cfbc08a63e0c73c7333f28d4a40835

jterapin:
Thanks for the PR! I'll merge it and will release a new version of the gem sometime today.

PR #124, opened 2024-06-05
No longer rely on caller_runs for backpressure in sqs active job executor

### LINKED ISSUE #123: Clean shutdown procedure is not invoked when the main thread participates in thread pool execution
`Concurrent::ThreadPoolExecutor` is [initialized](https://github.com/aws/aws-sdk-rails/blob/3c4e4828715a26154a918a5583e08f4f7adebb73/lib/aws/rails/sqs_active_job/executor.rb#L15) with `fallback_policy: :caller_runs`
This means that the work [will be executed in the thread of the caller](https://ruby-concurrency.github.io/concurrent-ruby/1.1.8/file.thread_pools.html), instead of being given to another thread in the pool when the queue of waiting work has reached the maximum size (**backpressure**) and no new threads can be created.
In such case all the exceptions are [rescued](https://github.com/ruby-concurrency/concurrent-ruby/blob/b5e090b9da95c63cd006721ae6a7e594ec78512b/lib/concurrent-ruby/concurrent/executor/ruby_thread_pool_executor.rb#L361) in the `RubyThreadPoolExecutor#run_task`.
This means that `Aws::Rails::SqsActiveJob::Interrupt#run` [rescue block](https://github.com/aws/aws-sdk-rails/blob/f05437103a51dc52c41a4b349d3499580b83a677/lib/aws/rails/sqs_active_job/poller.rb#L57) is not triggered and clean shutdown procedure is not invoked.

alextwoods:
Changes the `fallback_policy` on the executor to `:abort` instead of using `caller_runs`.  This will raise an `Concurrent::RejectedExecutionError` when posting to the executor when it does not have capacity.  It then uses a [Concurrent:Event](https://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/Event.html) to block the calling thread until the next task completes.

*Issue #, if available:*
Fixes #123 

*Description of changes:*

### DIFF
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -23,6 +23,30 @@ module Aws
         end
 
         def execute(message)
+          post_task(message)
+        rescue Concurrent::RejectedExecutionError
+          # no capacity, wait for a task to complete
+          @task_complete.reset
+          @task_complete.wait
+          retry
+        end
+
+        def shutdown(timeout = nil)
+          @executor.shutdown
+          clean_shutdown = @executor.wait_for_termination(timeout)
+          if clean_shutdown
+            @logger.info 'Clean shutdown complete.  All executing jobs finished.'
+          else
+            @logger.info "Timeout (#{timeout}) exceeded.  Some jobs may not have " \
+                         'finished cleanly.  Unfinished jobs will not be removed from ' \
+                         'the queue and can be ru-run once their visibility timeout ' \
+                         'passes.'
+          end
+        end
+
+        private
+
+        def post_task(message)
           @executor.post(message) do |message|
             job = JobRunner.new(message)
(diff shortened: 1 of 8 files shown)


lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -12,18 +12,19 @@ class Executor
...
         }.freeze
 
         def initialize(options = {})
           @executor = Concurrent::ThreadPoolExecutor.new(DEFAULTS.merge(options))
           @retry_standard_errors = options[:retry_standard_errors]
           @logger = options[:logger] || ActiveSupport::Logger.new($stdout)
+          @task_complete = Concurrent::Event.new
         end
 
         def execute(message)
```
mullermp, line 25: I think you can use implicit `begin` here but it will require another method. You possibly should refactor this out.

mullermp (approved):
Needs a test I think.

PR #126, opened 2024-06-06
Use Concurrent.available_processor_count to set default thread pool max threads

### LINKED ISSUE #125: Consider using `Concurrent.available_processor_count` to set default threads count
In the containerized environment, [calling Concurrent.processor_count](https://github.com/aws/aws-sdk-rails/blob/f05437103a51dc52c41a4b349d3499580b83a677/lib/aws/rails/sqs_active_job/poller.rb#L16) gives us host machine processor count.
[The proper question to ask is how many CPU cores are usable](https://github.com/ruby-concurrency/concurrent-ruby/pull/1038) and now we can do that by calling `Integer(Concurrent.available_processor_count)`.
It would require [sticking](https://github.com/aws/aws-sdk-rails/blob/f05437103a51dc52c41a4b349d3499580b83a677/aws-sdk-rails.gemspec#L29C36-L29C48) to the [`'concurrent-ruby', '>= 1.3.1'`](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/CHANGELOG.md#release-v131-29-may-2024).
It should help avoid exceeding database connection pools and slowing down job execution time for many people using K8s.

alextwoods:
Use Concurrent.available_processor_count to set default thread pool max threads

*Issue #, if available:*
Fixes #125 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Unreleased Changes
 ------------------
 
+* Issue - Use `Concurrent.available_processor_count` to set default thread pool max threads (#125).
 * Issue - No longer rely on caller_runs for backpressure in sqs active job executor (#123).
 
 3.12.0 (2024-04-02)
--- a/lib/aws/rails/sqs_active_job/executor.rb
+++ b/lib/aws/rails/sqs_active_job/executor.rb
@@ -9,7 +9,7 @@ module Aws
       class Executor
         DEFAULTS = {
           min_threads: 0,
-          max_threads: Concurrent.processor_count,
+          max_threads: Concurrent.available_processor_count.to_i,
           auto_terminate: true,
           idletime: 60, # 1 minute
           fallback_policy: :abort # Concurrent::RejectedExecutionError must be handled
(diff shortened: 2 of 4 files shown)


jterapin (approved):
Nice!

lib/aws/rails/sqs_active_job/executor.rb
```diff
@@ -9,7 +9,7 @@ module SqsActiveJob
       class Executor
         DEFAULTS = {
           min_threads: 0,
-          max_threads: Concurrent.processor_count,
+          max_threads: Integer(Concurrent.available_processor_count || Concurrent.processor_count),
```
mullermp, line 12: Can either of these ever be nil? Why is Integer wrapping processor_count too?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 Unreleased Changes
 ------------------
 
+* Issue - Use `Concurrent.available_processor_count` to set default thread pool max threads (#125).
 * Issue - No longer rely on caller_runs for backpressure in sqs active job executor (#123).
```
mullermp, line 5: wrap caller_runs in backticks

lib/aws/rails/sqs_active_job/executor.rb
alextwoods: `Concurrent.available_processor_count` can be nil, but `Concurrent.processor_count` will not.  `Concurrent.available_processor_count` returns a float, which needs to be converted to an Integer for max_threads.

PR #127, opened 2024-06-20
Add Action mailbox SES

ssunday:
~Hot mess of a PR. Very much not done.~

Basically, this is attempting to merge in https://github.com/bobf/action_mailbox_amazon_ingress 

**- Need to confirm license changes for other folks**
- ~Need to finish switching to rspec~
- ~Need to figure out how to get active storage migration in dummy app~
- ~Add docs~
- ~Test it actually works~

~I'm not sure when I'll get to wrapping this up/working on this again...so anyone is free to bring it to the finish line.~


mullermp:
Thanks for starting this effort. We do support older Rails 5.2. We are not opposed to a major version bump that has a new minimum rails of 7 if it makes things simpler. As mentioned, we're not opposed to rspec either, let's just make sure that the existing test cases are covered. This will probably be a large change and review for sure. Is there a way to do this in manageable bite sized pieces? Maybe we do the rspec conversion of aws-sdk-rails as is, without the action mailbox changes, and then we do that one?

ssunday:
@mullermp agreed with breaking it out.

mullermp:
There are a lot of conflicts here. Let me know how you want to proceed before making sweeping changes.

ssunday:
@mullermp I'm going to rebase and just get the important mailbox stuff in (and its tests). I'm working on it!

ssunday:
not working yet, routing isn't functional..like the routes aren't being mounted.

mullermp:
@bobf Would you be interested in reviewing and helping get this into aws-sdk-rails?

ssunday:
Got it working. I will need to test on an app this works e2e, and we also need to confirm everyone is OK with the license.

aws-sdk-rails.gemspec
```diff
@@ -24,12 +24,20 @@ Gem::Specification.new do |spec|
...
   spec.add_dependency('aws-sdk-sqs', '~> 1', '>= 1.56.0') # for ActiveJob
+  spec.add_dependency('aws-sdk-sns', '~> 1.75') # for ActionMailbox
 
   spec.add_dependency('aws-sessionstore-dynamodb', '~> 2') # includes DynamoDB
   spec.add_dependency('concurrent-ruby', '>= 1.3.1') # Utilities for concurrent processing
-  spec.add_dependency('railties', '>= 5.2.0') # encrypted credentials
+  spec.add_dependency('railties', '>= 7.0.0') # encrypted credentials
+  spec.add_dependency('actionmailbox', '>= 7.0.0')
+
+  spec.add_runtime_dependency('aws-sdk-s3', '~> 1.152')
```
ssunday, line 34: I'm not sure if we should just add it is a dependency? Or if SNS needs to be runtime too

mullermp:
You should add it into sample_app and then a section on the README for how to test with it.

aws-sdk-rails.gemspec
mullermp: We can add it as a runtime dependency here.

aws-sdk-rails.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
   spec.license       = 'Apache-2.0'
 
   spec.require_paths = ['lib']
-  spec.files += Dir['lib/**/*', 'bin/*']
+  spec.files += Dir['lib/**/*', 'bin/*', 'app/**/*']
```
ssunday, line 17: assuming this is correct

spec/dummy/config/boot.rb
```diff
@@ -0,0 +1 @@
+require 'bundler/setup'
```
ssunday, line 1: some of the stuff in dummy is so I could get bin/rails commands working since I was losing my mind

aws-sdk-rails.gemspec
ssunday: Actually...no, I don't think that can work due to the initializer and mounting

ssunday:
@mullermp yep, I need to pull in the docs from the other repo.

ssunday:
@mullermp I'm not sure how to locally test with the sample_app (I mean in this use case specifically).

bobf:
> @bobf Would you be interested in reviewing and helping get this into aws-sdk-rails?

Yep, sure. I'll try to find some time this week.

.github/workflows/ci.yml
```diff
@@ -24,21 +24,18 @@ jobs:
...
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
-
-      - uses: actions/checkout@v3
-      - name: Install gems
-        run: bundle install --gemfile=gemfiles/${{ matrix.gemfile }}
-
+          rubygems: latest
```
ssunday, line 36: https://github.com/ruby/setup-ruby?tab=readme-ov-file#rubygems

.github/workflows/ci.yml
```diff
@@ -24,21 +24,18 @@ jobs:
             ruby: 2.7
           - gemfile: rails-main.gemfile
             ruby: 3.0
-
+    env:
+      BUNDLE_GEMFILE: gemfiles/${{ matrix.gemfile }}
```
ssunday, line 28: We'll see if this change cleans it up or not 

https://github.com/ruby/setup-ruby?tab=readme-ov-file#matrix-of-gemfiles

ssunday:
I have another speculative fix for the tests/database issue—I'll push it up once I separately do an e2e test on a project that these changes even work 🤡

bobf:
@ssunday @mullermp I've scanned through the code (admittedly not very thoroughly, but I did read everything) and this all looks good to me.

Is there any specific feedback you want from me, or anything I can do to help this along ?

@ssunday I've lost track of how many times I've thanked you for all this hard work but it still seems insufficient - it's great to see somebody taking charge of this.

ssunday:
@bobf Early next week I'm going to get this code e2e tested on a real project so that'll test if it actually works in the wild. Once that's done, I just have to get tests working (which I have a speculative change I didn't push up since...I didn't want to push if the e2e test proved things needed more work anyway). So I've got those two things handled/planned for next week.

The thing I'm not sure of is licensing change since other people worked on this and I'm just doing the 'glue' work of getting it into this repo. Do they need to sign-off on it being into this repo with a diff license or is that not necessary with it from an MIT-origin repo? 

Thank _you_ for like getting this whole ball rolling and actually doing the hard work haha. This integration has greatly improved QOL on two of my projects. Having it as a proper package is just great for long term maintenance...

bobf:
I believe I'm able to authorise a licence change. As long as the change is to a similar license (Apache, BSD, etc. then all good with me).

ssunday:
Looks like it's working e2e.

mullermp:
Were you able to test locally and is it possibly to add a "how to test" entry and integrate with sample app in this repo?

ssunday:
@mullermp Well, the code I brought over comes with rspec examples that test the code. I can add those rspec integrations to the sample app....BUT then I'd need to convert the sample app to rspec from mini test. 

Beyond that, there's not really anything I can think of—honestly, most of the work in setup is on AWS. You configure two things in the app and then it just works assuming AWS SES/SNS is setup properly.

spec/aws/rails/action_mailbox/requests/rspec_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      it 'delivers an email to inbox' do
+        amazon_ingress_deliver_email(mail: Mail.new(to: 'user@example.com'))
+        expect(ActionMailbox::InboundEmail.last.mail.recipients).to eql ['user@example.com']
+      end
+    end
+
+    describe 'when unrecognized topic' do
+      let(:topic) { 'topic:arn:other' }
+      it 'renders 401 Unauthorized' do
+        amazon_ingress_deliver_email
```
ssunday, line 48: testing the helper methods @mullermp and the docs also talk about it

bobf:
Agree on this - the gem just expects a JSON payload with the correct structure and signature, so I don't know how else to test it other than with fixtures.

When I've run this locally I've had to use ngrok which is a bit of a pain.

mullermp:
That seems fine. Maybe an entry in that testing file about why NOT to have that and how fixtures are sufficient. We often do those manual / integration tests when we make "uneasy" changes just to be certain.

mullermp:
I'll be able to do a deeper review of this tomorrow or Thursday. Thank you both for your efforts.

ssunday:
@mullermp Cool! I'll hold on doing anything until after you do the deeper review. Thank you so much!

ssunday:
trying to fix tests still 😦

spec/dummy/db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2018_09_17_164000) do
+ActiveRecord::Schema[7.0].define(version: 2018_09_17_164000) do
```
ssunday, line 13: i did this pseudo manually...trying to run it for 7.0 caused bigints -> ints which we do not want

spec/aws/rails/action_mailbox/requests/inbound_email_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    post inbound_email_url, params: JSON.parse(fixture_for(:inbound_email, type: :json)), as: :json
+
+    expect(response).to have_http_status(:no_content)
+    expect(ActionMailbox::InboundEmail.count).to eql 1
+  end
+
+  it 'receives an inbound email with data in s3' do
+    s3_email = fixture_for(:s3_email, type: :txt)
+
+    s3_client = Aws::S3::Client.new(stub_responses: true)
```
ssunday, line 26: the old way might have been causing issues in other specs, I don't know

ssunday:
Yay, tests passing...except for rails main and jruby. The rails man ones look unrelated or general regression—can we run CI against main, perhaps? 

Also, side question-why do we run CI against rails-main? It's not an officially supported release and unstable. It seems like you would not run it/support it and only support cut releases.

ssunday:
I don't know how to debug jruby/never worked with it, so I'd appreciate help

ssunday:
I'm wondering if sqlite3 is not compatible with jruby period/doesn't play well (read something that said that).

spec/dummy/config/database.yml
```diff
@@ -0,0 +1,3 @@
+test:
+  adapter: sqlite3
```
mullermp, line 2: I think with JRuby you will need to conditionally add https://rubygems.org/gems/jdbc-sqlite3 from the Gemfile

spec/dummy/config/environments/test.rb
```diff
@@ -0,0 +1,10 @@
+Rails.application.configure do
+  config.active_storage.service = :test
+  config.action_mailbox.ingress = :amazon
+  config.action_mailbox.amazon.subscribed_topics = %w(
+    arn:aws:sns:eu-west-1:111111111111:example-topic
+    arn:aws:sns:eu-west-1:111111111111:recognized-topic
```
mullermp, line 6: You don't have to do this, but we typically use 012345678910 as account ID because it's easier to confirm typos of 12 digits.

spec/dummy/db/schema.rb
mullermp: What do you mean? I'd prefer if this file was generated and not a "pet" if that makes sense. Can you regenerate it and let's start from there?

spec/test_helper.rb
```diff
@@ -15,3 +16,11 @@ def deliverable(options = {})
     mail(options)
   end
 end
+
+def fixture_for(name, type:)
+  File.read(File.join('spec', 'fixtures', type.to_s, "#{name}.#{type}"))
+end
+
+RSpec.configure do |config|
+  config.use_transactional_fixtures = true
```
mullermp, line 25: Why is this necessary? (Just wondering)

Rakefile
```diff
@@ -10,13 +13,76 @@ Dir.glob('**/*.rake').each do |task_file|
...
-  t.libs << 'test'
-  t.pattern = 'test/**/*_test.rb'
-  t.warning = false
+RuboCop::RakeTask.new
+
+task default: :spec
+task 'release:test' => :spec
+
+PRIVATE_KEY = 'spec/fixtures/pem/privatekey.pem'
+CERTIFICATE = 'spec/fixtures/pem/certificate.pem'
```
mullermp, line 22: Is it possible we can generate these on the fly for when we run the tests? AWS Security does not like embedded / fixtures for certificates and keys.

mullermp (commented):
Nice work so far!

spec/test_helper.rb
ssunday: because entries are updated/created and they needed to be cleared out

ssunday: OG gem had a before to destroy all inbound emails

aws-sdk-rails.gemspec
```diff
@@ -23,13 +23,21 @@ Gem::Specification.new do |spec|
   # Require these versions for user_agent_framework configs
   spec.add_dependency('aws-sdk-ses', '~> 1', '>= 1.50.0') # for ActionMailer
   spec.add_dependency('aws-sdk-sesv2', '~> 1', '>= 1.34.0') # for ActionMailer
+  spec.add_dependency('aws-sdk-sns', '~> 1.75') # for ActionMailbox
   spec.add_dependency('aws-sdk-sqs', '~> 1', '>= 1.56.0') # for ActiveJob
 
+  spec.add_dependency('actionmailbox', '>= 7.0.0')
```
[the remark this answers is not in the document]
ssunday: correct

lib/aws/rails/action_mailbox/sns_notification.rb
```diff
@@ -0,0 +1,102 @@
...
+    module ActionMailbox
+      class SnsNotification
+        class MessageContentError < StandardError; end
+
+        def initialize(request_body)
+          @request_body = request_body
+        end
+
+        def subscription_confirmed?
+          confirmation_response.code&.start_with?('2')
```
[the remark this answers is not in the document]
ssunday: I don't know the meaning of this

spec/aws/rails/action_mailbox/requests/inbound_email_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'aws-sdk-s3'
```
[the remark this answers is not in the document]
ssunday: it was blowing up since I was ref'ing the client directly in the test

I think so?

spec/dummy/config/database.yml
ssunday: I'll try to figure out how to do that

README.md
```diff
@@ -163,6 +164,93 @@ You simply need to configure Rails to use it in your environment configuration:
...
+
+2. [Configure the SNS topic for SES or for the S3 action](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-action-sns.html) to send notifications to +/rails/action_mailbox/amazon/inbound_emails+. For example, if your website is hosted at https://www.example.com then configure _SNS_ to publish the _SES_ notification topic to this _HTTP_ endpoint: https://example.com/rails/action_mailbox/amazon/inbound_emails
+
+#### Rails
+
+1. Configure _ActionMailbox_ to accept emails from Amazon SES:
+
+```
+# config/environments/production.rb
+config.action_mailbox.ingress = :amazon
```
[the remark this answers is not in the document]
ssunday: sure

Rakefile
ssunday: I didn't write this code, not sure, can look into it one day

aws-sdk-rails.gemspec
ssunday: I don't know how this works, to be honest, this is the deepest I've gone with rails gems

aws-sdk-rails.gemspec
ssunday: I mean, we can make it a regular dependency no problem

lib/aws/rails/action_mailbox/rspec.rb
```diff
@@ -0,0 +1,51 @@
...
+          post email.url,
+               params: email.params,
+               headers: email.headers,
+               as: :json
+        end
+
+        private
+
+        def message_verifier(subscription_confirmation)
+          instance_double(Aws::SNS::MessageVerifier, authentic?: subscription_confirmation.authentic?)
```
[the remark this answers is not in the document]
ssunday: not sure why this was done this way

spec/dummy/db/schema.rb
ssunday: Previous commits **did** try to run migrations in tests, but it was running into issues with certain rails versions. Hence changing it to schema. I can go back to the run migrations way but I'm pretty sure that is going to cause other problems.

.rubocop.yml
```diff
@@ -29,10 +31,14 @@ Style/GlobalVars:
...
 
 Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*.rb'
 
 Style/HashSyntax:
   EnforcedShorthandSyntax: never
+
+Style/Documentation:
+  Enabled: false
```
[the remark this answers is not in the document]
ssunday: **bunch** of stuff, adding exclusions

aws-sdk-rails.gemspec
```diff
@@ -23,13 +23,21 @@ Gem::Specification.new do |spec|
...
   spec.add_dependency('aws-sessionstore-dynamodb', '~> 2') # includes DynamoDB
   spec.add_dependency('concurrent-ruby', '>= 1.3.1') # Utilities for concurrent processing
-  spec.add_dependency('railties', '>= 5.2.0') # encrypted credentials
+  spec.add_dependency('railties', '>= 7.0.0') # encrypted credentials
+
+  spec.add_runtime_dependency('aws-sdk-s3', '~> 1.152')
+
+  spec.add_development_dependency('pry')
```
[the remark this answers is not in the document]
ssunday: Well. I guess you can say either way https://docs.rubocop.org/rubocop/cops_gemspec.html#gemspecdevelopmentdependencies

Can we commit to one way LOL

.rubocop.yml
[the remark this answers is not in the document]
mullermp: If it's api private, use `# @api private` and it will be counted as documentation. If it's public, we need real comments.

aws-sdk-rails.gemspec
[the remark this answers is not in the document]
mullermp: Yeah, I would prefer the default, of Gemfile here. gemspec would have only required dependencies.

aws-sdk-rails.gemspec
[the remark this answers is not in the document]
mullermp: I think if railties is loaded, likely rails is loaded right? Debating if we should even care to add this.

lib/aws/rails/action_mailbox/rspec.rb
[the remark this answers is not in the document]
mullermp: We can ask @bobf for a review again with my comments.

lib/aws/rails/action_mailbox/sns_notification.rb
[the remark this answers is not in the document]
mullermp: @bobf

spec/dummy/config/database.yml
mullermp: In the Gemfile (since that should have development dependencies), we can do
```
if defined?(JRUBY_VERSION)
  gem 'jdbc-sqlite3'
else
  gem 'sqlite3'
end
```

or something like this.

Rakefile
mullermp: Perhaps @bobf can help or provide insight on why this was needed.

README.md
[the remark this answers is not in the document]
ssunday: The only issue is that it'll be a breaking change for people trying to convert to this from the other gem AND I'd have to redo e2e tests

.rubocop.yml
```diff
@@ -3,15 +3,17 @@ inherit_from: .rubocop_todo.yml
...
     - 'tasks/release/**/*'
     - 'sample_app/**/*'
     - 'spec/dummy/**/*'
     - 'spec/fixtures/**/*'
     - 'spec/fixtures/**/*'
+    - 'db/**/*'
 
 Gemspec/DevelopmentDependencies:
-  Enabled: false
+  EnforcedStyle: Gemfile
```
ssunday, line 16: @mullermp added this so it's enforced

README.md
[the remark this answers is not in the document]
ssunday: I might be able to leave the route as is...tbd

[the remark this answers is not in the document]
ssunday: (nvm, your previous contradict says to convert to ses)

aws-sdk-rails.gemspec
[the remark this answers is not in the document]
ssunday: as someone who works on an app who pulls in rails gems piecemeal...you never know~~~

lib/aws/rails/action_mailbox/sns_notification.rb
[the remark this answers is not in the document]
ssunday: OH. I think it's saying the status code is in a 200 range

README.md
```diff
@@ -18,20 +18,14 @@ gem 'aws-sdk-rails', '~> 3'
...
 
+* `aws-sdk-s3`
 * `aws-sdk-ses`
 * `aws-sdk-sesv2`
 * `aws-sdk-sqs`
+* `aws-sdk-sns`
 * `aws-record`
 * `aws-sessionstore-dynamodb`
 
-If you want to use other services (such as S3), you will still need to add them to your
```
ssunday, line 27: this wording probably needs to be retained, but another ex is needed. I'm not sure if we actually need it tho or if it's self explanatory

spec/dummy/db/schema.rb
ssunday: Added a new commit that reverts back—I think it'll break for rails mainline...which is breaking other parts of this gem. I still argue it shouldn't even be tested.

README.md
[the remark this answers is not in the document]
mullermp: I think it's fair for that breakage since there's one dependency to another. `:ses` is the right config here even if there's a little pain to move over, we should have it right to begin with.

spec/dummy/config/environments/test.rb
ssunday: (this relates to the fixtures, so it's tangled)

ssunday: TBD

ssunday:
Okay, migration change failed

README.md
mullermp: This comment is a general AWS SDK explanation and how aws-sdk-rails does not bring in all of the SDK. V2 SDK was "monolithic" such that gem aws-sdk would install everything. It's probably fine to remove at this point in v3's lifecycle.

ssunday:
oh yeah, I removed the migrations, duh, I'll fix 🤦🏻

CONTRIBUTING.md
```diff
@@ -58,11 +58,15 @@ request](https://help.github.com/articles/creating-a-pull-request/).
 
 To setup the repository:
 
-1. `bundle install`
+1. `BUNDLE_GEMFILE=gemfiles/rails-7.1.gemfile bundle install`
```
mullermp, line 61: This will get stale. What's wrong with main? I know sometimes it fails but that's ok - it gives us a heads up on what is changing sometimes.

.rubocop.yml
```diff
@@ -29,10 +31,19 @@ Style/GlobalVars:
...
 
 Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*.rb'
 
 Style/HashSyntax:
   EnforcedShorthandSyntax: never
+
+Style/Documentation:
+  Exclude:
```
mullermp, line 44: I think the `# @api private` is applicable for most if not all of these. It should satisfy rubocop and not have them be documented in yard.

aws-sdk-rails.gemspec
mullermp: The fact that it's in app doesn't matter, I think you can move the file as is into lib and strip off app/controllers and it's still fine.

.rubocop.yml
ssunday: I'm not changing code not related to this PR

spec/dummy/db/schema.rb
ssunday: (I have to fix more stuff 😦 )

ssunday:
The remaining requests are pretty large, not sure when I'll get to it. Anyone else is free to dive in and work on them 👍🏻 .

ssunday:
Actually, I'll work on the renaming of the routes (amazon -> ses)

aws-sdk-rails.gemspec
ssunday: I don't believe so due to autoloading. Based on another gem that add controller, this is OK https://github.com/ankane/ahoy/tree/master/app/controllers/ahoy

Rakefile
ssunday: Okay...so these update the fixtures in json with the proper values based on the content. (so if you change content, must run the rake task to update). 

It looks like the signing cert comes from AWS SNS publicly and it's used to do that. Pulling live seems like it'd be fragile/slower.

lib/aws/rails/action_mailbox/rspec.rb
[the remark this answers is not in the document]
ssunday: I think this does do that with how the opts are passed in. I think this is OK and flexible.

lib/aws/rails/action_mailbox/sns_notification.rb
```diff
@@ -0,0 +1,102 @@
...
+          @notification ||= JSON.parse(@request_body, symbolize_names: true)
+        rescue JSON::ParserError => e
+          Rails.logger.warn("Unable to parse SNS notification: #{e}")
+          nil
+        end
+
+        def s3_content
+          require 'aws-sdk-s3'
+
+          Aws::S3::Client
```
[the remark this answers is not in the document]
ssunday: Well, since this is called from a controller I think even if it's memoized it'd still be regenerated unless it's set at boot time, right, and then if it's set at boot time it would be instantiating for those not using it 🤔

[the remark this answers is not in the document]
ssunday: added the framework from my understanding 🤔

Rakefile
mullermp: Is there a public link for that? I'd rather get it from the public source even if it's slower. It shouldn't be fragile I don't think? We could always fetch on each run.

Gemfile
```diff
@@ -4,12 +4,23 @@ source 'https://rubygems.org'
...
+
 group :development do
   gem 'rubocop'
 end
 
 group :test do
   gem 'bcrypt'
+  gem 'rspec-rails'
+  if defined?(JRUBY_VERSION)
+    gem 'activerecord-jdbcsqlite3-adapter'
```
ssunday, line 19: this seems more apt https://github.com/jruby/activerecord-jdbc-adapter

Rakefile
ssunday: Ugh, it seems to be, but I don't know what the current one is. You can see how it refs it in the json fixtures themselves—like it returns a `SigningCertURL`

ssunday: I'd say fragile as it has like a UUID on the URL. 

https://docs.aws.amazon.com/sns/latest/dg/sns-verify-signature-of-message.html

Okay, here's what it says

ssunday:
```
Get the X509 certificate that Amazon SNS used to sign the message. The SigningCertURL value points to the location of the X509 certificate used to create the digital signature for the message. Retrieve the certificate from this location.
Extract the public key from the certificate. The public key from the certificate specified by SigningCertURL is used to verify the authenticity and integrity of the message.
```

I'm not sure this a concern. I think the cert is like arbitrary?

ssunday: Like we are stubbing out the call to get the pem with the one stored in the app.

Gemfile
ssunday: Okay, Jruby seems to have a psych issue, I believe the fix is just installing psych before the tests...TBD

ssunday: ...I can't replicate it locally

ssunday: advice appreciated 😓

ssunday: blowing up cache to see if that works

mullermp: I have no clue, but I agree on using the other gem.

ssunday: I tried using it, but it's still going haywire on bundle install

mullermp:
The main branch failures look related to the PR, because main in this repo passes just fine (and the other PR you saw). I think maybe the route is interfering, you can see by the stack trace. I can help take a look at some of the items here.

ssunday:
@mullermp oooh, I wonder if it's because of the transactional fixtures. I'll try w/o that and see how that goes. 

If you could look into the jruby stuff, that'd be great. I'm lost with it haha

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
```diff
@@ -4,7 +4,7 @@
 
 module ActiveJob
   module QueueAdapters
-    class AmazonSqsAdapter
+    class AmazonSqsAdapter < AbstractAdapter
```
ssunday, line 7: @mullermp https://github.com/rails/rails/blob/4fa56814f18fd3da49c83931fa773caa727d8096/activejob/lib/active_job/queue_adapters/abstract_adapter.rb#L14 see that enqueue method (it was missing and that's why it was blowing up in tests).

I _think_ the request spec tests (when using sqs driver) were actually testing the queue flow fully whereas before it wasn't so this isn't raised on main.

spec/dummy/config/environments/test.rb
```diff
@@ -0,0 +1,11 @@
+Rails.application.configure do
+  config.active_storage.service = :test
+  config.action_mailbox.ingress = :ses
+  config.action_mailbox.ses.subscribed_topics = %w(
+    arn:aws:sns:eu-west-1:012345678910:example-topic
+    arn:aws:sns:eu-west-1:012345678910:recognized-topic
+  )
+
+  config.action_dispatch.show_exceptions = :none
+  config.active_job.queue_adapter = :test
```
ssunday, line 10: I feel like we should have the queue adapter be test so mailbox can be tested in isolation from it—I hope the sqs tests aren't like silently not testing anything now

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
ssunday: no that doesn't make any sense (the test coverage thing), but it does make sense from a rails-main angle

ssunday: Well. Maybe it's the transaction wrapper _and_ the rails main changes. Like it wasn't in a transaction before...now it is. So it caught something that wasn't caught before. I'll verify by enabling transactions on main and seeing if it melts.

ssunday: Well. Maybe it's the transaction wrapper _and_ the rails main change? Like it wasn't in a transaction before/no db stuff...now it is?

lib/active_job/queue_adapters/amazon_sqs_adapter.rb
```diff
@@ -5,6 +5,10 @@
 module ActiveJob
   module QueueAdapters
     class AmazonSqsAdapter
+      def enqueue_after_transaction_commit?
```
ssunday, line 8: this is gonna be needed anyway for new-rails release

ssunday: https://github.com/rails/rails/pull/51426

mullermp:
Ok, I spent some time trying to figure out what was going on. The original jruby failure was because of the bundler cache I think, so I reverted that bit, then desperately tried a few things but I believe I got it all working. The JDBC adapters track Rails versions so I had to conditionally pull those, and in the case of Rails 7.1 it was never published for some reason (opened an issue on their repo) and for 7.2 we will have to skip this case now. Is it possible to avoid any kind of database in the dummy for adding this feature?

Rakefile
mullermp: I'll investigate this and get back to you.

lib/aws/rails/action_mailbox/sns_notification.rb
```diff
@@ -0,0 +1,103 @@
...
+        private
+
+        def notification
+          @notification ||= JSON.parse(@request_body, symbolize_names: true)
+        rescue JSON::ParserError => e
+          Rails.logger.warn("Unable to parse SNS notification: #{e}")
+          nil
+        end
+
+        def s3_content
```
mullermp, line 54: We should memoize the client in initialize of this class or a higher level class (maybe in the controller somehow?) and pass it through.

spec/aws/rails/action_mailbox/requests/inbound_email_spec.rb
mullermp: What was the old way being used here?

spec/dummy/config/environments/test.rb
mullermp: I'll look into this.

lib/aws/rails/action_mailbox/sns_notification.rb
ssunday: I am pretty sure if you memoize it in the controller, it's still regenerated each request, so it doesn't matter.

Maybe in an initializer somewhere?

ssunday:
@mullermp `Is it possible to avoid any kind of database in the dummy for adding this feature?` Well, then it's not really testing the inbound emails flow actually works. It feels like a massive hit to testing the feature. 

🤯 on the jruby nonsense. Great work on figuring it out. With jruby—is it...that popular? Do we need to be that concerned over thoroughly supporting it versus having the ideal situation for ruby plain?

ssunday:
Rephrasing: I think it's more important to have the database tests to verify this feature works in some cases than having parity of testing for all ruby versions (if that makes sense).

.github/workflows/ci.yml
```diff
@@ -16,28 +16,44 @@ jobs:
...
           bundler-cache: true
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+
       - name: Install gems
-        run: bundle install --gemfile=gemfiles/${{ matrix.gemfile }}
+        run: bundle install
+        env:
+          BUNDLE_GEMFILE: gemfiles/rails-${{ matrix.rails }}.gemfile
```
ssunday, line 46: was the workspace path not working?

Gemfile
```diff
@@ -4,12 +4,36 @@ source 'https://rubygems.org'
...
+
 group :development do
   gem 'rubocop'
 end
 
 group :test do
+  # JDBC versions track Rails versions
+  if ENV['RAILS_VERSION'] == '7.0'
+    gem 'activerecord-jdbcsqlite3-adapter', '~> 70.0', platform: :jruby
+  elsif ENV['RAILS_VERSION'] == '7.1'
```
ssunday, line 19: brutal, but it's green...

spec/dummy/config/application.rb
```diff
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
 require 'rails'
+require 'activerecord-jdbc-adapter' if defined? JRUBY_VERSION
```
ssunday, line 4: fascinating

ssunday:
@mullermp Okay, re-reading. Not sure OOP with your comments anymore. It looks like everything's green, so it seems like there actually isn't an issue with the database setup anymore?? It's just the memoization and maybe the fixtures?

.github/workflows/ci.yml
mullermp: Yeah, I think that was causing the issue? I can try again now that I got everything green and see what was breaking.

spec/dummy/config/application.rb
mullermp: I think this ended up being necessary. I couldn't reproduce many things locally.

ssunday: I feel that 🫂

ssunday:
@mullermp I can't think of anything else rn—I know there's a request to handle KMS encryption (https://github.com/bobf/action_mailbox_amazon_ingress/pull/8) but that can be a subsequent release.

I'll need to test e2e once all the last changes are made. I'll hold off until Alex reviews and changes are made based on that/the other stuff.

mullermp:
Sounds good. It would be great if we can get a sample_app test if possible and/or manual test instructions using sample_app.

spec/dummy/config/application.rb
mullermp: Yeah, it was necessary only for the latest version. There is a CI run below that you can see.

config/routes.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+Rails.application.routes.draw do
```
mullermp, line 3: Shouldn't this file be in the gemspec too?

lib/aws/rails/action_mailbox/rspec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'aws/rails/action_mailbox/rspec/email'
+require 'aws/rails/action_mailbox/rspec/subscription_confirmation'
+require 'aws-sdk-sns'
+
+module Aws
+  module Rails
+    module ActionMailbox
+      # @api private
```
mullermp, line 10: These are public test methods that we are delivering, right? Should probably document each method here.

config/routes.rb
```diff
@@ -0,0 +1,8 @@
...
+  scope '/rails/action_mailbox', module: 'action_mailbox/ingresses' do
+    post '/ses/inbound_emails' => 'ses/inbound_emails#create',
+         as: :rails_ses_inbound_emails
```
mullermp, line 6: Is the `rails` prefix convention or necessary? Would prefer to remove.

lib/aws/rails/action_mailbox/sns_notification.rb
mullermp: Yeah, in an initializer or railstie somehow and passed through. The client needs to be configurable through code.

spec/aws/rails/action_mailbox/requests/inbound_email_spec.rb
mullermp: Should we resolve this?

spec/dummy/db/migrate/20180917164000_create_action_mailbox_tables.rb
```diff
@@ -0,0 +1,19 @@
+class CreateActionMailboxTables < ActiveRecord::Migration[6.0]
```
mullermp, line 1: Should this be version 7?

config/routes.rb
ssunday: I'm pretty sure it's necessary to match action mailbox routes

ssunday: Like necessary in the sense of following action mailbox setup flow

spec/dummy/db/migrate/20180917164000_create_action_mailbox_tables.rb
ssunday: I think the problem was it melting supporting non-Rails 7 versions or something, it doesn't really matter

lib/aws/rails/action_mailbox/rspec.rb
ssunday: they are documented in the readme, but I can add them here

config/routes.rb
ssunday: ohh true.

ssunday:
`Sounds good. It would be great if we can get a sample_app test if possible and/or manual test instructions using sample_app.`

I don't really feel like that's going to be that useful. In terms of code, there's only like 2 config things you setup which are already documented. The rest is in AWS.

mullermp:
Sorry, we had discussed that (I forgot - my bad! - a lot going on)

ssunday:
@mullermp Absolutely no problem, very much understand!

With the certificate stuff, I think if you look at how the test code is using it, you'll see that the persisted certs are like...whatever. They can be regenerated by anyone and the fixtures we saved without issue. Like they aren't...real in the eyes of AWS. You can make one and regen test fixtures. Like they aren't supposed to be an actual cert from AWS, but in place of via stubbing. That's my take on it.

spec/dummy/db/migrate/20180917164000_create_action_mailbox_tables.rb
ssunday: Yeah. But I don't think we have to change the version anyway, like it works.

.github/workflows/ci.yml
ssunday: huh, weird

ssunday: not a big deal

spec/aws/rails/action_mailbox/requests/inbound_email_spec.rb
ssunday: Oh, it was like doing Aws[:s3] = blah blah. This way feels cleaner and maybe more inline with the other tests

mullermp:
Understood and I agree . AWS security flags this kind of stuff and will ultimately bug me about it. Let me just confirm with them that it's ok.

spec/aws/rails/action_mailbox/requests/inbound_email_spec.rb
ssunday: https://github.com/bobf/action_mailbox_amazon_ingress/blob/master/spec/requests/inbound_email_spec.rb#L23

config/routes.rb
ssunday: For reference, you can see the route structure: https://guides.rubyonrails.org/action_mailbox_basics.html

ssunday: I mean we want to follow the rails pattern*

lib/aws/rails/action_mailbox/rspec.rb
ssunday: neat!

ssunday:
rebased + force pushed

ssunday:
I think setting the dummy app action job driver to SQS will cause CI to fail—I'd like to make it test mode. It's failing locally.

lib/aws/rails/action_mailbox/sns_notification.rb
ssunday: The only issue is that it'll change how the code works—right now region is pulled from the notification rather than being static.

ssunday: Why does it need to be configurable? For like a MVP release?

mullermp:
>I think setting the dummy app action job driver to SQS will cause CI to fail—I'd like to make it test mode. It's failing locally.


I didn't see this happening when I ran locally

ssunday:
@mullermp were you running against rails main gemfile?

ssunday:
CI did fail in the way that I expected

ssunday:
@mullermp I think it's cause of the rails-main transaction thing with enqueing. It doesn't try to enqueue on non-rails-man, thus no error with the SQS driver. But with rails-man...it does so it blows up as we haven't stubbed sqs and it tries to call out to AWS.

I still say we switch to test mode for the dummy app.

lib/aws/rails/action_mailbox/sns_notification.rb
ssunday: We could make the region for the topic as config option then I can memoize it...it's another step, potentially, for setup, but it can be done.

ssunday: Okay, done....let me know if you like how I did it. I think it'll memoize properly.

lib/aws/rails/action_mailbox/s3_client.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'aws-sdk-s3'
+
+module Aws
+  module Rails
+    module ActionMailbox
+      # @api private
+      class S3Client
```
ssunday, line 9: I can change name...just did what was simplest lol

lib/aws/rails/action_mailbox/sns_notification.rb
```diff
@@ -0,0 +1,98 @@
...
+        def initialize(request_body)
+          @request_body = request_body
+        end
+
+        def subscription_confirmed?
+          (200..299).cover?(confirmation_response.code.to_i)
+        end
+
+        def verified?
+          Aws::SNS::MessageVerifier.new.authentic?(@request_body)
```
alextwoods, line 22: The MessageVerifier will cache pems - we may want to memoize it (similar to what was done for the S3 client) to take advantage of that and avoid downloading the same cert repeatedly.

lib/aws/rails/action_mailbox/sns_notification.rb
alextwoods: Looking at the example config in the README, it lists two queues in different regions, but configures the s3 client with only one region.  I think we may still have an issue with this memoization.  

Initializing clients is expensive (esp S3), so I agree we should be memoizing this somehow and agree that in general they should be configurable.  Maybe a method to retrieve/create an s3 client for a given region which would merge the `ses.s3_client_options`?

Something like:
```
def s3_client(region)
  @clients[region] ||= Aws::S3::Client.new(Rails.configuration.action_mailbox.ses.s3_client_options.merge(region: region)
end
```

For reference, the config I'm referring to is:
```
# config/environments/production.rb
config.action_mailbox.ses.subscribed_topics = %w(
  arn:aws:sns:eu-west-1:012345678910:example-topic-1
  arn:aws:sns:us-east-1:012345678910:example-topic-2
)
config.action_mailbox.ses.s3_client_options = { region: 'us-east-1' }
```

alextwoods (commented):
Overall looks good!

README.md
```diff
@@ -163,6 +157,94 @@ You simply need to configure Rails to use it in your environment configuration:
...
+### Testing
+
+#### RSpec
+
+Two _RSpec_ _request spec_ helpers are provided to facilitate testing _Amazon SNS/SES_ notifications in your application:
+
+* `action_mailbox_ses_deliver_subscription_confirmation`
+* `action_mailbox_ses_deliver_email`
+
+Include the `ActionMailboxAmazonIngress::RSpec` extension in your tests:
```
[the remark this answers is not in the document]
ssunday: Ah, yes, didn't fix it here but in the other docs

lib/aws/rails/action_mailbox/sns_notification.rb
ssunday: How about we don't allow multiple subscribed topics for now and let that be an additional release (and someone can figure out how to easily do the s3 client stuff)?

I don't use multiple so I don't care 😁

Rakefile
```diff
@@ -3,16 +3,85 @@
...
+  MessageId
+  Subject
+  SubscribeURL
+  Timestamp
+  Token
+  TopicArn
+  Type
+].freeze
+
+file PRIVATE_KEY do |t|
```
alextwoods, line 34: If we can't move this logic into the test runners themselves, can we move all of these test fixture/cert tasks to a separate file?

Rakefile
alextwoods: I believe whats being done here is fairly reasonable since we are generating random test certificates and then updating the signatures in the fixtures to match.

However - It seems like this is something that could be done instead as an rspec before(:all) and then avoids any need for running these tasks when tests are changed or storing test certs.

ssunday: Hmm. Well, it's saving changes to the fixture files. Or are you saying we don't save the files, like merge it into the data returned? 

Seems a little more complicated

lib/aws/rails/action_mailbox/sns_message_verifier.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+require 'aws-sdk-sns'
+
+module Aws
+  module Rails
+    module ActionMailbox
+      # @api private
+      class SnsMessageVerifier
```
ssunday, line 9: memoized verifier

lib/aws/rails/action_mailbox/sns_message_verifier.rb
```diff
@@ -0,0 +1,18 @@
...
+        class << self
+          def client
+            @client ||= Aws::SNS::MessageVerifier.new
```
alextwoods, line 12: nit: This isn't really a client, maybe just @verifier?

README.md
```diff
@@ -176,14 +176,11 @@ config.action_mailer.delivery_method = :ses # or :sesv2
 config.action_mailbox.ingress = :ses
 ```
 
-2. Configure which _SNS_ topics will be accepted and what region the emails will be stored in when using S3 (plus any other desired options):
+2. Configure which _SNS_ topic will be accepted and what region the emails will be stored in when using S3 (plus any other desired options):
```
alextwoods, line 179: Feel free to disregard, but it may be worth mentioning that the `s3_client_options[:region]` does not need to match the region of the subscribed_topic, but must match the region for the configured s3 bucket.  (I find this allowed mismatch confusing personally....)

mullermp:
I'll be able to help with this tomorrow and hopefully we can get this merged by early next week.

lib/aws/rails/action_mailbox/sns_message_verifier.rb
ssunday: true, I wasn't thinking too hard about naming tbh. I'll change

ssunday:
@mullermp I need to test e2e again with it—I should be able to do that today after I make the feedback changes.

ssunday:
@mullermp Well, I don't have to, but it seems like it's a good idea since the amount of changes since my last e2e change.

mullermp:
An e2e test from your side would be great.

Rakefile
ssunday: I think the test runner comment has been handled.

Why would we move into it a separate file if it's a rake task?

lib/aws/rails/action_mailbox/sns_notification.rb
mullermp: I think we can try to help with this.

mullermp:
Sorry I've been a bit MIA - I haven't forgotten about this. I may not get to this properly until next week.

mullermp:
It's basically there other than Alex's comments. I can do another pass potentially this week or next week at the latest.

ssunday:
@mullermp The rakefile comment seems like it shouldn't be a blocker since it's test/cleanup related. Still, I'd rather wait 😅

mullermp (approved):
Nice work! I think this is finished. I agree that subscribed topics (plural) can come later and any fixture refactors can also be done later. I've moved them into a separate rake task for organization. I think a last end to end review would be appreciated and we can try to publish this or next week.

ssunday:
E2E went fine so far! 👍🏻

alextwoods (approved):

alextwoods:
The description still has "Need to confirm license changes for other folks".  Has that been done?  What are the license changes/issues?

mullermp:
I believe so, bobf said it was fine to be Apache which we have here.

mullermp:
Thank you @ssunday for your hard work and contribution and @bobf for initial leg work.

bobf:
Great work, everyone. @ssunday Let me know if I can send a case of beer or something similar your way, I really do appreciate you getting this all wrapped up.

ssunday:
@bobf Thank _you_! The existence of this code in general has greatly improved the QOL projects and infrastructure... less 💦 for me! Now it's in a nice cozy home lol.

timlawrenz:
thanks @ssunday

PR #128, opened 2024-06-24
[TEST] Convert to rspec for non-sample app specs

ssunday:
Convert to rspec for tests (broken out from https://github.com/aws/aws-sdk-rails/pull/127 for sanity) 


spec/test_helper.rb
```diff
@@ -1,13 +1,8 @@
...
 
-require 'minitest/autorun'
-require 'minitest/spec'
-require 'minitest/unit'
-require 'rspec/mocks/minitest_integration'
-require 'rspec/expectations/minitest_integration'
-
 ENV['RAILS_ENV'] = 'test'
 require_relative 'dummy/config/application'
+require 'rspec/rails'
```
ssunday, line 5: necessary order change

ssunday:
Is there a reason to keep supporting end of life-d rails/ruby versions...? I feel like they should just be cut out/not officially supported. 

Will do the rubocop fixes.

mullermp:
We can definitely consider a major version bump and drop old versions of Rails/Ruby. The SDK supports 2.5+ and in general we try to keep support where we can. We should keep the same matrix paradigm where we test every minor version of rails with ruby.

mullermp:
I think it's reasonable to start at Rails 7, and Ruby 2.7 (minimum of Rails 7)

ssunday:
@mullermp 👍🏻 Works for me. I'll make that change to the matrix and do the other things

mullermp:
We should update the CI, README, VERSION, and gemspec possibly in yet another PR... I'd be happy to merge that quickly to unblock this one.

ssunday:
https://github.com/aws/aws-sdk-rails/pull/129 let's see if I did it right

aws-sdk-rails.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('concurrent-ruby', '>= 1.3.1') # Utilities for concurrent processing
   spec.add_dependency('railties', '>= 5.2.0') # encrypted credentials
   spec.add_development_dependency('rails')
+  spec.add_development_dependency('rspec-rails', '~> 6.1')
```
mullermp, line 32: Should we always pull the latest here?

mullermp (commented):
Next I think it's just rubocop and see what CI does.

aws-sdk-rails.gemspec
mullermp: It's up to you - I have not used rspec-rails. Would major versions break the code here? I assume if it does it would be simple to constrain back.

ssunday: actually, let's target it since clearly it doesn't work for all rails versions

ssunday: though, we should probably target rails as well

aws-sdk-rails.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
...
 
-  spec.required_ruby_version = '>= 2.5'
+  spec.required_ruby_version = '>= 2.7'
```
ssunday, line 34: FYI @mullermp

aws-sdk-rails.gemspec
ssunday: https://github.com/rspec/rspec-rails/blob/v6.1.3/Changelog.md#600--2022-10-10 on breaking changes for ex

mullermp:
I don't know if I merged this prematurely or not. I see in the other PR actually that you've converted actual tests to use RSpec.describe and new syntax! Let me know if that's the case.

ssunday:
@mullermp I can just do that in the other PR/keep changes, when I was working on it there I didn't realize I didn't have to to MVP get it done lolol.

ssunday:
@mullermp this is good to get merged, IMO.

PR #132, opened 2024-07-18
Require action mailbox engine by default

mullermp:


mullermp:
@ssunday

jterapin (approved):

ssunday (approved):

lib/aws/rails/action_mailbox/engine.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'action_mailbox/engine'
```
ssunday, line 3: well. Would this auto set it up for people who don't use action mailbox?

ssunday: or auto 'enable' mailbox for their rails app even though it's not in use?

ssunday: I approved maybe pre-maturely, I'm not actually sure what this does 😅

mullermp: `aws/rails/action_mailbox/engine` is being loaded regardless. If this is not loaded, it blows up.
```
irb(main):001> require 'rails'
=> true
irb(main):002> require 'aws-sdk-rails'
/Users/mamuller/.gem/ruby/3.3.3/gems/railties-7.1.3.4/lib/rails/railtie/configuration.rb:109:in `method_missing': undefined method `action_mailbox' for an instance of Rails::Engine::Configuration (NoMethodError)
```

ssunday: OOooooh, got it

Issue #133, opened 2024-07-19
V4 - Usage without ActiveStorage and ActiveMailbox enabled


davidwessman:
Hello!

We are using the gem for sending email via SES, but we do not have ActionMailbox or ActiveStorage loaded in the app.
On gem version 4.0.1 we therefore cannot load the app due to the error:

```ruby
Couldn't find Active Storage configuration in /home/runner/work/repo/config/storage.yml
```

Our `application.rb` starts like:

```ruby
require_relative "boot"

require "rails"
# Pick the frameworks you want:
require "active_model/railtie"
require "active_job/railtie"
require "active_record/railtie"
# require "active_storage/engine"
require "action_controller/railtie"
require "action_mailer/railtie"
# require "action_mailbox/engine"
# require "action_text/engine"
require "action_view/railtie"
require "action_cable/engine"
require "sprockets/railtie"
require "rails/test_unit/railtie"
```

Is there no way to use this gem unless you have ActionMailbox and ActiveStorage enabled?

mullermp:
Thanks for opening an issue. I think that's the case currently and I'm open to suggestions for how to fix that. Was that also the case for ActiveStorage in version 3?

davidwessman:
> Thanks for opening an issue. I think that's the case currently and I'm open to suggestions for how to fix that. Was that also the case for ActiveStorage in version 3?

We are currently using version 3 with no problems.

mullermp:
I'm not able to reproduce this when I create a new rails app. I tried with both `require 'rails/all'` and with the requires you pasted above. I'm assuming your app was pretty old but then have upgraded it many times? What happens if you add a blank storage.yml file in config? Or at very least, define one for your test environment?

I'm assuming you ran into this because of the new ActionMailbox feature which uses ActiveStorage under the hood.

davidwessman:
> I'm not able to reproduce this when I create a new rails app. I tried with both `require 'rails/all'` and with the requires you pasted above. I'm assuming your app was pretty old but then have upgraded it many times? What happens if you add a blank storage.yml file in config? Or at very least, define one for your test environment?
> 
> 
> 
> I'm assuming you ran into this because of the new ActionMailbox feature which uses ActiveStorage under the hood.

I am on vacation until mid August and will try to stay away from my work computer. Will get back to you.

benjaminwols:
I've run into the same issue. We also don't require `active_storage` and `active_mailbox` in our app.

[...]

PR #134, opened 2024-07-20
Consolidate the loading of modules related to AWS SQS ActiveJob

kakubin:
partially related: #133

We are using the gem just for processing async job on ActiveJob.
Since we don't need any other functionality of the gem, we would like a way to use it easily like the following.

```ruby
gem 'aws-sdk-rails', require: false

require 'aws/rails/sqs_active_job' # load modules just about worker

module MyApplication
  class Application < Rails::Application
    config.active_job.queue_adapter = :sqs
  end
end
```

### DIFF
--- a/spec/aws/rails/sqs_active_job_spec.rb
+++ b/spec/aws/rails/sqs_active_job_spec.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+module Aws
+  module Rails
+    describe SqsActiveJob do
+      describe '.config' do
+        before { Aws::Rails::SqsActiveJob.instance_variable_set(:@config, nil) }
+
+        it 'creates and returns configuration' do
+          expect(Aws::Rails::SqsActiveJob::Configuration).to receive(:new).and_call_original
+          expect(Aws::Rails::SqsActiveJob.config).to be_a Aws::Rails::SqsActiveJob::Configuration
+        end
+
+        it 'creates config only once' do
+          expect(Aws::Rails::SqsActiveJob::Configuration).to receive(:new).once.and_call_original
+          # call twice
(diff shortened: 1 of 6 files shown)


mullermp:
Thanks. This seems fine. Would you be able to also move the associated class config tests out of the configuration spec and into a new spec file that matches where they exist now?

kakubin:
> Would you be able to also move the associated class config tests out of the configuration spec and into a new spec file that matches where they exist now?

Done

kakubin:
To be honest, since other dependencies like aws-sdk-s3 are not needed in my case, I would like AWS SQS ActiveJob to be separated into a gem and managed similarly to aws-sdk-ruby.

spec/aws/rails/sqs_active_job_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+          end
+
+          expect(Aws::Rails::SqsActiveJob.config).to have_attributes(
+            visibility_timeout: 360,
+            excluded_deduplication_keys: contain_exactly('job_class', 'job_id')
+          )
+        end
+      end
+
+      describe '#queue_url_for' do
```
mullermp, line 38: This one still belongs in Configuration.

mullermp (approved):

mullermp:
Hm. I wonder if we can drop direct dependency and require them to be loaded instead. I'll think about that.

spec/aws/rails/sqs_active_job_spec.rb
kakubin: oups... sorry!
I fixed

mullermp:
Thanks for your fix. I'm open to any PRs for making other features optional and not having other SDK gems as direct dependencies.

PR #135, opened 2024-07-22
Check if action mailbox config is defined

### LINKED ISSUE #133: V4 - Usage without ActiveStorage and ActiveMailbox enabled

mullermp:
Potentially fixes #133

### DIFF
--- a/lib/aws/rails/action_mailbox/engine.rb
+++ b/lib/aws/rails/action_mailbox/engine.rb
@@ -5,10 +5,12 @@ module Aws
     module ActionMailbox
       # @api private
       class Engine < ::Rails::Engine
-        initializer 'aws-sdk-rails.mount_engine' do |app|
-          app.config.action_mailbox.ses = ActiveSupport::OrderedOptions.new
-          app.config.action_mailbox.ses.s3_client_options ||= {}
+        if config.respond_to?(:action_mailbox)
+          config.action_mailbox.ses = ActiveSupport::OrderedOptions.new
+          config.action_mailbox.ses.s3_client_options ||= {}
+        end
 
+        initializer 'aws-sdk-rails.mount_engine' do |app|
           app.routes.append do
             mount Engine => '/'
           end
--- a/lib/aws/rails/action_mailbox/engine.rb
+++ b/lib/aws/rails/action_mailbox/engine.rb
@@ -5,14 +5,14 @@ module Aws
     module ActionMailbox
       # @api private
       class Engine < ::Rails::Engine
-        if config.respond_to?(:action_mailbox)
+        if defined?(ActionMailbox::Engine)
           config.action_mailbox.ses = ActiveSupport::OrderedOptions.new
           config.action_mailbox.ses.s3_client_options ||= {}
-        end
 
-        initializer 'aws-sdk-rails.mount_engine' do |app|
-          app.routes.append do
-            mount Engine => '/'
+          initializer 'aws-sdk-rails.mount_engine' do |app|
+            app.routes.append do
+              mount Engine => '/'
+            end
           end
         end
       end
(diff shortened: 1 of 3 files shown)


mullermp:
@ssunday @bobf Sorry to bother - but do you think this will fix #133? Would this change work in your application?

I think there's something smelly about this as it is currently. I tried using `ActiveSupport.on_load(:action_mailbox)` but to no avail. I also think maybe this engine should not be loaded all the time unless action mailbox is required. Should this also live in a railtie? Let me know what you all think.

ssunday:
@mullermp No idea. This seems like it makes sense, but I don't know.

ssunday:
It's a low-key problem with this gem is that it's providing all sorts of functionality even if you don't want it. Like it's pulling in aws gems that you may not need.

mullermp:
We made the decision a few years ago to bundle all the features because it's 1) easier for us to manage (develop/release) and 2) more "all in one" and "works out of the box" like rails is itself.

I think we want to keep this pattern but I agree we can probably not bring in SDK gems as dependencies but rather check if they are loaded and enable those features as such.

lib/aws/rails/action_mailbox/engine.rb
```diff
@@ -1,14 +1,14 @@
...
       # @api private
       class Engine < ::Rails::Engine
-        config.action_mailbox.ses = ActiveSupport::OrderedOptions.new
-        config.action_mailbox.ses.s3_client_options ||= {}
+        if config.respond_to?(:action_mailbox)
+          config.action_mailbox.ses = ActiveSupport::OrderedOptions.new
+          config.action_mailbox.ses.s3_client_options ||= {}
+        end
 
         initializer 'aws-sdk-rails.mount_engine' do |app|
```
alextwoods, line 13: Should this also be inside the if block?

lib/aws/rails/action_mailbox/engine.rb
```diff
@@ -1,14 +1,14 @@
...
-require 'action_mailbox/engine'
-
 module Aws
   module Rails
     module ActionMailbox
       # @api private
       class Engine < ::Rails::Engine
-        config.action_mailbox.ses = ActiveSupport::OrderedOptions.new
-        config.action_mailbox.ses.s3_client_options ||= {}
+        if config.respond_to?(:action_mailbox)
```
alextwoods, line 8: Should we also check that ActionMailbox is loaded? (ie, check if the module `ActionMailbox` is defined?)

mullermp: ActionMailbox can be loaded (i.e. `require 'action_mailbox'`) but this can still fail, it really needs `action_mailbox/engine`.

lib/aws/rails/action_mailbox/engine.rb
mullermp: I guess it could be. But I think there's a larger problem about where this actually belongs

alextwoods (approved):

PR #137, opened 2024-07-31
Revert action mailer config validation

### LINKED ISSUE #136: Sending Action mailer emails from different regions in the same application instance
Before version 4.0.2 (specifically https://github.com/aws/aws-sdk-rails/pull/131) it was possible to configure multiple Action mailer delivery methods with different configurations, such as:

```ruby
Aws::Rails.add_action_mailer_delivery_method(:east, credentials: creds, region: "us-east-1")
Aws::Rails.add_action_mailer_delivery_method(:west, credentials: creds, region: "us-west-1")
```

The result was a fallback to using `ses` (v1) for the two delivery methods `east` and `west`. You could then decide which delivery method, and thus region, to use when sending an email.

Because of the stricter validations added in 4.0.2, only `ses` and `sesv2` are now accepted as the first `name` argument.

Since defining two delivery methods this way is no longer possible I assume it is not (or has ever been?) the recommended way of setting up multiple delivery methods backed by SES.

However, from the existing documentation it is not obvious to me what the correct way would be, or if there is a completely different approach that is better suited for this.

Any suggestions?

mullermp:
Fixes #136

It looks like name was overloaded for adding additional mailers of the same class. We could consider adding a "v2" delivery method instead of a major version bump, but I'd rather the method signature `add_action_mailer_delivery_method(name, mailer, client_options = {})`. I can't add mailer as a default param to the end because of keyword arguments potentially being used as client options.

### DIFF
--- a/lib/aws/rails/railtie.rb
+++ b/lib/aws/rails/railtie.rb
@@ -9,7 +9,7 @@ module Aws
                   before: :load_config_initializers do
         # Initialization Actions
         Aws::Rails.use_rails_encrypted_credentials
-        Aws::Rails.add_action_mailer_delivery_method(:ses)
+        Aws::Rails.add_action_mailer_delivery_method
         Aws::Rails.add_action_mailer_delivery_method(:sesv2)
         Aws::Rails.log_to_rails_logger
       end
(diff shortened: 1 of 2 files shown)


lib/aws/rails/railtie.rb
```diff
@@ -32,15 +32,14 @@ class Railtie < ::Rails::Railtie
     #   register, either :ses or :sesv2.
     # @param [Hash] client_options The options you wish to pass on to the
     #   Aws::SES[V2]::Client initialization method.
-    def self.add_action_mailer_delivery_method(name, client_options = {})
+    def self.add_action_mailer_delivery_method(name = :ses, client_options = {})
+      # TODO: on the next major version, add a "mailer" param to this method
```
alextwoods, line 36: Maybe add a link to the specific Issue for context when we get to this.

alextwoods (approved):
LGTM - and agree this is probably the right fix within this MV and free form name + configured class (ses/sesv2) is probably the right long term (new mv) fix.

PR #140, opened 2024-09-16
SDK Eager loading

alextwoods:
*Issue #, if available:*

*Description of changes:*
Adds in eager loading for the SDK to optimize the start up/require times.  See: https://github.com/aws/aws-sdk-ruby/pull/3105 for more context and in particular see [(revereted) commit](https://github.com/aws/aws-sdk-ruby/pull/3105/commits/44fd76db259a1fcf0360cfbf1380c29f96ec794b#diff-b6240d08ff9163bd4ab7228022ca912b81d68f07d62abed19757f8c2370c34b4) that added railties to each gem with eager_load.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Feature - Add SDK eager loading to optimize load times.  See: https://github.com/aws/aws-sdk-ruby/pull/3105.
+
 4.0.3 (2024-07-31)
 ------------------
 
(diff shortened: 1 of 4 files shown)


mullermp (approved):
Consider a test for eager load (maybe confirm that it's called)

Schwad (approved):
Yeah maybe something like this 


```ruby
require 'test_helper'

class EagerLoadingTest < ActiveSupport::TestCase
  def test_aws_sdk_rails_eager_loading
    Rails.application.config.eager_load = true

    Aws::Rails.expects(:eager_load!)
    Rails.application.eager_load!
  end
end

```

Thanks for extracting this!

PR #144, opened 2024-10-01
Sqs active job as own gem

ckhsponge:
*Issue #, 142*

*Description of changes:*

Moved sqs-active-job to a top level directory and created it's own gemspec.


ckhsponge:
Tests are working yet, the dummy environment needs to be unwound from other dependencies.

mullermp:
Thanks. I'll get back to you in this. We will want to think about gem naming and plan out work to move features out of the rails gem.

ckhsponge:
Sounds good. I'll do some real world of testing of this meanwhile.

mullermp:
Thanks. It may be a bit before I get to this. I am thinking maybe it shouldn't live in this GitHub if it's not exclusively for rails. We have a gem called aws-sessionstore-dynamodb that works for rack applications, and then we have a rails layer for it. I am thinking we would want an aws-activejob-sqs gem and repository. I am thinking whether or not Rails loading checks should be done in both cases, removed from aws-sdk-rails. E.g. If Rails was loaded, then insert a railtie or look for config.

ckhsponge:
@mullermp A completely seperate gem does seem like a decent and logical way to go. However, ActiveJob is part of the Rails project so keeping AWS support for ActiveJob close to Rails support could make sense as well.

I agree that the conditional Rails loading checks aren't great. There's only 2 of them and I suspect there may be a way to remove them.

aws-sdk-rails.gemspec
```diff
@@ -32,5 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency('concurrent-ruby', '~> 1.3', '>= 1.3.1') # Utilities for concurrent processing
   spec.add_dependency('railties', '>= 7.0.0') # Minimum supported Rails version
 
+  spec.add_dependency('sqs-active-job', version)
```
alextwoods, line 35: I think we potentially would want this to be an open ended dependency

aws-sdk-rails.gemspec
alextwoods, line 1: Does `spec.executables = ['aws_sqs_active_job']` still work with the gem split out?

sqs-active-job/lib/sqs_active_job.rb
```diff
@@ -15,6 +16,8 @@ module Rails
     #
     # SQS-based queuing backend for Active Job.
     module SqsActiveJob
+      VERSION = File.read(File.expand_path('../../VERSION', __dir__)).strip
```
alextwoods, line 19: I think we'd probably want to start versioning this separately

aws-sdk-rails.gemspec
alextwoods: I think we can also remove sqs and concurrent-ruby from the dependencies here.

sqs-active-job/lib/sqs_active_job/poller.rb
```diff
@@ -22,17 +22,11 @@ class Poller
...
-
-        def set_environment
-          @environment = @options[:environment] || ENV['APP_ENV'] || ENV['RAILS_ENV'] || ENV['RACK_ENV'] || 'development'
+          @options[:queue] ||= :default
+          # please set environment and boot rails BEFORE initializing this poller
         end
 
         def run
-          # exit 0
-          boot_rails
```
alextwoods, line 35: I'm not sure the best way to do this, but what about trying to require rails and running boot_rails only when its available? something like:
```
begin
  require 'rails'
  boot_rails
rescue LoadError
 # no boot
end
```

alextwoods (commented):
This is looking good - great work and thanks for starting this.  I think its a great idea to break this out as an independent gem!

sqs-active-job/lib/sqs_active_job/poller.rb
ckhsponge: Thanks for reviewing.

I moved the Rails booting to the executable which I think is a better place for it. The poller doesn't need to control what environment it's polling in. If you really want the poller to do booting we could split the poll method to boot and then call poll_without_booting_rails. This new method poll_without_booting_rails could then be used by non-rails environments that want to do polling. Other environments can easily create their own executables if desired.

mullermp:
Work has been moved to https://github.com/aws/aws-activejob-sqs-ruby/pull/1

PR #149, opened 2024-10-24
Modularize Aws::Record

mullermp:
* Checks for Aws::Record before adding things
* Fixes the tests (were previously not running)
* General code cleanup
* README and usage improvements


lib/generators/aws_record/base.rb
```diff
@@ -1,213 +0,0 @@
-# frozen_string_literal: true
-
-require 'rails/generators'
-require_relative 'generated_attribute'
-require_relative 'secondary_index'
-
-module AwsRecord
-  module Generators
-    class Base < Rails::Generators::NamedBase
```
alextwoods, line 9: I think I'm slightly confused about this - Did this functionality move?

alextwoods: Edit: oh nvm, I see in model_generator.rb

lib/generators/aws_record/model/model_generator.rb
```diff
@@ -1,13 +1,56 @@
...
+require 'rails/generators'
+
+require_relative 'generated_attribute'
+require_relative 'secondary_index'
 
 module AwsRecord
   module Generators
-    class ModelGenerator < Base
+    # Generates an Aws::Record model and table configuration.
+    class ModelGenerator < Rails::Generators::NamedBase
```
alextwoods, line 11: I think I'm still missing something - where does this get required from?

test/tasks/aws_record/test_table_config.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'aws-record'
+
+module ModelTableConfig
```
alextwoods, line 5: What does this do/how it is called/tested?

alextwoods (commented):
Nice - overall looks good, just a few questions on areas I don't quite understand.

lib/generators/aws_record/model/model_generator.rb
mullermp: It's loaded automatically because of its subclass. Just like the session store generators.

lib/generators/aws_record/base.rb
mullermp: I removed the base class because it doesn't make sense, and put all of the functionality inside the generator.

test/tasks/aws_record/test_table_config.rb
mullermp: The migrate rake script looks for all modules of this name in the db/table_config directory and calls this method. In the migrate rake test class, we mock that search to return this file, and when the method is called, the mock is verified to call the migrate method.

alextwoods: Ahh that makes sense.  It might be worth adding a comment somewhere about that, I couldn't trace the magic going on.

alextwoods (approved):

PR #152, opened 2024-10-28
Use AbstractStore for DynamoDbSession

mullermp:
Uses a delegator to RackMiddleware in aws-sessionstore-dynamodb as a long term fix to missing functionality.

Also bumps the dependency to ~> 3 (which has not been released yet)


mullermp:
CI is expected to be broken until aws-sessionstore-dynamodb ~> 3 is released

CHANGELOG.md
```diff
@@ -1,10 +1,14 @@
 Unreleased Changes
 ------------------
```
alextwoods, line 3: Nit: Mention dropping of support for rails 7.0 here.

lib/action_dispatch/session/dynamo_db_store.rb
```diff
@@ -16,28 +16,39 @@ module Session
...
-      # generate_sid in Compatibility takes precedence over our generate_sid.
-      # Compatibility is needed so that the request is an ActionDispatch::Request.
-      CompatibilityWithoutSid = Compatibility.dup
-      CompatibilityWithoutSid.remove_method(:generate_sid)
-      include CompatibilityWithoutSid
-
-      include StaleSessionCheck
-      include SessionObject
-
+    class DynamoDbStore < ActionDispatch::Session::AbstractStore
```
alextwoods, line 19: Nice - I like this approach!

CHANGELOG.md
mullermp: It says
* Feature - Depend on `aws-sessionstore-dynamodb ~> 3` which depends on `rack ~> 3` and is not supported by Rails `7.0`.

Issue #16, opened 2017-06-30
Seems to not do RFC-2047 encoding


jeznag:
Hi, I'm using this gem to send emails via SES and am finding that it breaks when I use email addresses with non-ASCII characters. e.g.  用户@例子.广告

`Aws::SES::Errors::InvalidParameterValue (Local address contains control or whitespace):`

From the docs, it seems that the email address must be encoded according to RFC-2047. Is there a way to handle this case easily?

awood45:
I need some more background to help you out, as I haven't heard of this issue before. Can you point me to the documentation you're referring to? I can go from there to try and reproduce this and point you to a workaround.

awood45:
Nevermind, I [found the relevant documentation](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/send-email-raw.html).

Unfortunately, it does appear that the SDK is behavior correctly with regards to what the API expects. Essentially, that's an ASCII string, with that encoding already done. You'd need to sanitize your email inputs to be compliant before sending. Are there issues that make this especially difficult?

jeznag:
Ok I'm not really sure how to do that. Can't seem to find any docs on using RFC-2047 encoding with action mailer. Linking to StackOverflow question in case someone comes across this issue via Google: https://stackoverflow.com/questions/44876345/action-mailer-with-international-email-addresses-rfc-2047

mullermp:
I'm going to assume that this is solved?

PR #156, opened 2024-11-05
EB SQS Docker IP check overhaul to correct 403 issue (use remote IP + remote addr)

ssunday:
*Issue #, if available:*

https://github.com/aws/aws-sdk-rails/issues/154

*Description of changes:*

~~Taken code from https://github.com/active-elastic-job/active-elastic-job/blob/master/lib/active_elastic_job/rack/sqs_message_consumer.rb to resolve issue—seems to be the old IP check in this gem doesn't work with AL2023?~~

Issue was that we needed to check `remote_addr`


ssunday:
I'm not sure if the unreleased changes to the old code would have worked, but this seems to.

lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
```diff
@@ -96,23 +96,13 @@ def in_docker_container_with_cgroup2?
         File.exist?('/proc/self/mountinfo') && File.read('/proc/self/mountinfo') =~ %r{/docker/containers/}
       end
 
-      def default_gw_ips
-        default_gw_ips = ['172.17.0.1']
```
ssunday, line 100: This just seems to resolve in giving me two entries of `172.17.0.1` on the AL2023 test, which seems legit. Maybe the issue is the non-remote ip/addr usage? IIRC when I ran this, the debug logger had the IP as something else. I can add more debug statements to prove what's going on (this is very weird).

ssunday:
```
D, [2024-11-05T17:13:44.713584 #16] DEBUG -- : aws-sdk-rails middleware detected call from Elastic Beanstalk SQS Daemon.
D, [2024-11-05T17:13:44.713881 #16] DEBUG -- : SQSD request detected from IP: 127.0.0.1.
D, [2024-11-05T17:13:44.713926 #16] DEBUG -- : SQSD request detected from Remote IP: 127.0.0.1.
```

Maybe it needs to be remote_addr 🤯 . Trying again

ssunday: if this is the root cause—checking `request.ip` does not work—then I wonder if using the OG code is fine, it just needs to check against r**emote_ip and remote_addr** instead. I assume it would since it seemed like it was calculating properly. 

I'll make those changes once I confirm it really must be remote_addr

ssunday: `D, [2024-11-05T17:41:38.678283 #16] DEBUG -- : SQSD request detected from Remote ADDR: 172.17.0.1.`

HA! There it is.

mullermp:
Thanks. I'll take a look. Sorry for the delays. I'm trying to set up an elastic beanstalk rails website but the step by step guide that is published doesn't produce something that works.

lib/aws/rails/middleware/ebs_sqs_active_job_middleware.rb
ssunday: Updated code!

spec/aws/rails/middleware/ebs_sqs_active_job_middleware_spec.rb
```diff
@@ -4,193 +4,287 @@
 require_relative 'elastic_beanstalk_job'
 require_relative 'elastic_beanstalk_periodic_task'
 
-module Aws
-  module Rails
-    describe EbsSqsActiveJobMiddleware do
-      # Simple mock Rack app that always returns 200
-      let(:mock_rack_app) { ->(_) { [200, { 'Content-Type' => 'text/plain' }, ['OK']] } }
+describe Aws::Rails::EbsSqsActiveJobMiddleware do
```
ssunday, line 7: massively re-did the test suite, let me know if I missed any cases 🤯 or it could be simplified

ssunday:
@mullermp no worries, I just figured out the actual root cause and pushed up proper fixing. 

What you'd really need is both a ruby-rails EB stack and a Docker EB stack that happens to run rails 🤯

ssunday:
@mullermp I haven't tested this round of updates in reality, but it was the remote_addr lack of checking that seemed to be the core issue, based on debugging

alextwoods:
I've finally managed to get this setup and tested - looks like the change is working!  Thank you for all the debugging work on this!

alextwoods (approved):

PR #159, opened 2024-11-07
Relocate ActionMailer and deprecate its railtie method

mullermp:
This change moves the ActionMailer mailers to https://github.com/aws/aws-actionmailer-ses-ruby/pull/1 and changes their namespace and class names. The railtie method that has caused issues is also removed. With this approach, customers are advised to configure their own delivery method using whatever class and name they want to, as well as providing client options. This change also updates the testing guide for this feature.


mullermp:
@admtnner This is where I did the refactoring work - I ended up removing automatic registration. I think I'll keep that bit regardless, but it sounds like you want this in a separate gem, too.

sample-app/README.md
```diff
@@ -87,4 +87,30 @@ To override changes, change this app's Gemfile to use the local path.
...
+
+`TestMailer` implemented SES and SESv2 mailer methods.
+
+`MailerController` (and routes) were added to send mail.
+
+`config/application.rb` added `require "action_mailer/railtie"`.
+
+Delivery methods are configured in `config/initializers/action_mailer.rb`.
+
+**Important**: The email address in SES must be verified.
```
jterapin, line 110: Could we do a "call out" box to highlight importance? 
> **Important**: The email address in SES must be verified.

Gemfile
```diff
@@ -4,6 +4,8 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'aws-actionmailer-ses', git: 'https://github.com/aws/aws-actionmailer-ses-ruby', branch: 'init'
```
alextwoods, line 7: why branch init here? Is this just temporary for testing until we merge the other PR?

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ Unreleased Changes
 
 * Issue - Do not skip autoload modules for `Aws::Rails.instrument_sdk_operations`.
 
+* Issue - Add deprecation warning to `Aws::Rails.add_action_mailer_delivery_method` to instead use `ActionMailer::Base.add_delivery_method`.
+
+* Issue - New namespace and class names for SES and SESV2 mailers. Existing namespace has temporarily been kept for backward compatibility. These now live in the `aws-actionmailer-ses` gem.
```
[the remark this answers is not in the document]
mullermp: It's already getting feature bumped so I didn't care too much. I'll fix.

Gemfile
mullermp: Yeah this was temporary. I'm going to again keep this here until we release aws-sdk-rails and all four gems have been published.

README.md
```diff
@@ -202,14 +202,24 @@ use the provided Rake task:
 rake dynamo_db:session_store:clean
 ```
 
-## Amazon Simple Email Service (SES) as an ActionMailer Delivery Method
+## ActionMailer delivery with Amazon Simple Email Service
```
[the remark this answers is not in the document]
mullermp: That's the plan but I wasn't sure whether to do that now or later (trying to avoid confusion) because the other readme says to install the gem which doesn't exist. For now I just duplicated the info and will delete this in one sweep.

PR #162, opened 2024-11-13
Relocate ActionMailbox

mullermp:
Relocates the ActionMailbox SES feature to https://github.com/aws/aws-actionmailbox-ses-ruby/pull/1


mullermp:
@ssunday fyi

Gemfile
```diff
@@ -6,6 +6,7 @@ gemspec
 
 gem 'aws-actiondispatch-dynamodb', git: 'https://github.com/aws/aws-actiondispatch-dynamodb-ruby'
 gem 'aws-actionmailer-ses', git: 'https://github.com/aws/aws-actionmailer-ses-ruby'
+gem 'aws-actionmailer-ses', git: 'https://github.com/aws/aws-actionmailer-ses-ruby', branch: 'init'
```
ssunday, line 9: ugh, shouldn't this be MAILBOX?

ssunday (changes requested):

Gemfile
mullermp: Yeah. Copy paste error... fixed it

sample-app/config/initializers/action_mailer.rb
```diff
@@ -1,3 +1,3 @@
 options = {}
-ActionMailer::Base.add_delivery_method :ses, Aws::ActionMailer::SESMailer, **options
-ActionMailer::Base.add_delivery_method :ses_v2, Aws::ActionMailer::SESV2Mailer, **options
+ActionMailer::Base.add_delivery_method :ses, Aws::ActionMailer::SES::Mailer, **options
```
ssunday, line 2: isn't this an unrelated to mailbox change

ssunday (commented):
Sqlite can be killed from the gemfiles/X, no?

sample-app/config/initializers/action_mailer.rb
mullermp: It's in the sample-app. A change I made for the other gem. I just forgot to change it here, and when testing ActionMailbox, I got an unrelated error.

mullermp:
>Sqlite can be killed from the gemfiles/X, no?

Yes. I can do this now. My original plan was to just do a sweeping clean up after I got all of the functionality out.

alextwoods (approved):

mullermp:
@ssunday All approved from our side. I think you can actually test this if you pull in aws-sdk-rails from github/branch in your gemfile, this gem and the other gems the same way. It should all still work. Otherwise, we can defer that and if you want to give approves that's fine too.

PR #164, opened 2024-11-17
Prepare aws-sdk-rails 4.2 modularization

mullermp:
This change uses the modular gems and fixes up the CI to make the release tools work.


mullermp:
@ssunday If you are able, could you please pull this branch, and test it in your application? Can you also specify which features you are using? This branch should pull in the newly published modular gems (there are 5 of them). Everything in this major version should be backwards compatible. If it's not, please let me know what sharp edges you run into and I can put out some fixes. I plan to make breaking changes on the modular gems ~> 1 (they are published as 0.1.0 now) and on aws-sdk-rails ~> 5.

mullermp:
Any bravehearts are welcome to also do some testing..

ssunday:
Haven't done any testing, but it seems like one day this gem can go away if the rails specific logic becomes optional in each gems (so they are not rails specific but have the rails feature set). Like it's weird to have to pull in the omnibus gem if you only want like the sqs functionality, but if you use EB you also need this...etc. 

Will try to test it out soon on a project!

mullermp:
The middleware can technically be used with any adapter, not just sqs. That's why it belongs here imo. Also, some generic features like encrypted credentials, notifications, rails logger as SDK logger, etc, belong here and did not make sense to split out.

I appreciate volunteering to test! I did manual tests and unit tests on my own but it doesn't hurt to be sure.

ssunday:
Makes sense!

alextwoods (approved):

ssunday:
Been looking good so far

PR #166, opened 2024-11-20
Prepare aws-sdk-rails 5

mullermp:
Major version bumps aws-sdk-rails plus some additional changes:

* Drops dependencies on modular gems.
* Removes `Aws::Rails.action_mailer_delivery_method`.
* Removes old action mailbox require path.
* Removes old symlinked namespaces.
* ActiveSupport Notification plugin added by default.
* Moved initializers into their own initialization method and set appropriate ordering.
* Do not execute ActiveJob with cron unless path is root.


aws-sdk-rails.gemspec
```diff
@@ -13,14 +13,9 @@ Gem::Specification.new do |spec|
   spec.license       = 'Apache-2.0'
   spec.files         = Dir['LICENSE.txt', 'CHANGELOG.md', 'VERSION', 'lib/**/*']
 
-  # These will be removed in aws-sdk-rails ~> 5
-  spec.add_dependency('aws-actiondispatch-dynamodb', '~> 0')
```
alextwoods, line 17: nice

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -31,37 +31,42 @@ def call(env)
...
+        rescue NameError
+          internal_error_response
+        end
+
+        def _execute_periodic_task(job_name)
+          job = job_name.constantize.new
+          job.perform_now
+        rescue NameError => e
+          @logger.error("Periodic task #{job_name} could not resolve to an Active Job class " \
+                        '- check the spelling in cron.yaml.')
```
alextwoods, line 67: Maybe mention change the url if this should not be processed as a job?

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -31,37 +31,42 @@ def call(env)
...
-          begin
-            ::ActiveJob::Base.execute(job)
-          rescue NameError => e
-            @logger.error("Job #{job_name} could not resolve to a class that inherits from Active Job.")
-            @logger.error("Error: #{e}")
-            return internal_error_response
-          end
-
+          _execute_job(job, job_name)
           [200, { 'Content-Type' => 'text/plain' }, ["Successfully ran job #{job_name}."]]
```
alextwoods, line 39: Nit, if we have `internal_error_response` should we also have a method for successful respones?

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -31,37 +31,42 @@ def call(env)
...
-          begin
-            job = job_name.constantize.new
-            job.perform_now
-          rescue NameError => e
-            @logger.error("Periodic task #{job_name} could not resolve to an Active Job class - check the spelling in cron.yaml.")
-            @logger.error("Error: #{e}.")
-            return internal_error_response
-          end
-
+          _execute_periodic_task(job_name)
```
alextwoods, line 56: Where is the path being checked?

lib/aws/rails/railtie.rb
```diff
@@ -46,65 +36,37 @@ class Railtie < ::Rails::Railtie
...
-      # TODO: remove this method in aws-sdk-rails ~> 5
-      ActiveSupport.on_load(:action_mailer) do
-        if name == :sesv2
-          add_delivery_method(name, Aws::Rails::Sesv2Mailer, client_options)
-        else
-          add_delivery_method(name, Aws::Rails::SesMailer, client_options)
+      # Add ActiveSupport Notifications instrumentation to AWS SDK client operations.
+      # Each operation will produce an event with a name `<operation>.<service>.aws`.
+      # For example, S3's put_object has an event name of: put_object.S3.aws
+      initializer 'aws-sdk-rails.instrument-sdk-operations', after: :load_active_support do
```
alextwoods, line 43: I'm not sure this should be on by default - I think theres a cost of the extra plugin + processing the notifications.  Additionally, might be order sensitive - this will add plugins only to services that have already been required

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
mullermp: The messages are different, there is not much re-use there.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
mullermp: Oh, I made an oopsie, but tests passed.

mullermp: It's checked in `periodic_task?` which only invokes the job by name. Otherwise, cron will just call a route, and the route can do whatever (out of scope for this middleware)

lib/aws/rails/railtie.rb
mullermp: I debated this - load_active_support is very early in the chain, executed 2nd https://guides.rubyonrails.org/configuring.html#initializers. I think even if you add this thing as a custom initializer (which we recommended already) it is effectively being done the same way, while the application is initializing. It would have already required users to require the sdk before doing so.

I think the performance impact is fairly small. On the ActiveSupport side, it's a block being called by wrapper, and on the plugin side, it's just an additional plugin which we add frequently.

So in short, I'm not sure what to do about it. We could swing the other way and have nothing with a railtie, but I think it's nice to have these things automatically registered.

alextwoods: yeah, if you think its low cost, then that seems fine to keep by default.

alextwoods (approved):

Issue #167, opened 2024-11-06
nginx timeout for jobs that take more than 60 sec (default nginx timeout)


januszm:
I'm currently migrating from [Shoryuken](https://github.com/ruby-shoryuken/shoryuken) task queue (a [Sidekiq](https://github.com/sidekiq/sidekiq) clone that uses SQS as queue store) to AWS SDK Rails, and I've encountered a strange error. 
I mention Sidekiq because it's a de-facto state of the art task queue for Ruby world.
For longer jobs that require e.g. 5-10 minutes to complete, I get a timeout, and the message returns to the queue. It looks like the POST request to the application server is waiting for the task to complete, which is unacceptable for longer jobs, because some can take up to several hours. Am I doing something wrong, or is this software designed so that sqs daemon, when submitting a task to be executed using the POST method, waits for it to **complete**? I expected it to only wait for the task to be "**accepted**" for async execution, not for it to complete.

alextwoods:
The fix has been merged and released in 5.1.0.

januszm:
Amaz(on)ing! Thank you.

PR #168, opened 2024-12-03
Support async job processing in EBS SQSD middleware

alextwoods:
*Issue #, if available:*
#167 

*Description of changes:*
Adds new ENV configuration to opt-in to async job processing in EBS SQSD middleware.

Async job processing is done using a thread pool executor.  When there is capacity, the job is queued and a 200 OK is returned immediately.  When there is no capacity, the middleware returns a 429 (too many requests) which will cause the SQSD to NOT delete the message, leaving it in the queue for future processing. 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
-          cache-version: 0912834390128
+          bundler: latest
 
       - name: Test
         run: bundle exec rake spec
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,6 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
-          bundler: latest
 
       - name: Test
         run: bundle exec rake spec
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          cache-version: 0912834390128
 
       - name: Test
         run: bundle exec rake spec
--- a/README.md
+++ b/README.md
@@ -151,11 +151,17 @@ the `AWS_PROCESS_BEANSTALK_WORKER_THREADS` environment variable.
 
 When there is no additional capacity to execute a task, the middleware
 returns a 429 (too many requests) response which will result in the 
-sqsd NOT deleting the message.  The mesagge will be retried again once its
+sqsd NOT deleting the message.  The message will be retried again once its
+visibility timeout is reached.
+
+Periodic (scheduled) tasks will also be run asynchronously in the same way.
+Elastic beanstalk queues a message for the periodic task and if there is
+no capacity to execute the task, it will be retried again once the message's
 visibility timeout is reached.
 
 #### Periodic (scheduled) jobs
-Periodic (scheduled) jobs are also supported with this approach. Elastic
+[Periodic (scheduled) tasks](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features-managing-env-tiers.html#worker-periodictasks)
+are also supported with this approach. Elastic
 Beanstalk workers support the addition of a `cron.yaml` file in the application
 root to configure this. You can call your jobs from your controller actions
 or if you name your cron job the same as your job class and set the URL to
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ the `AWS_PROCESS_BEANSTALK_WORKER_REQUESTS` environment variable to `true` in
 the worker environment configuration. The
 [SQS Daemon](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features-managing-env-tiers.html#worker-daemon)
 running on the worker sends messages as a POST request to `http://localhost/`.
-The aws-sdk-rails middleware will forward each request and parameters to their
+The ElasticBeanstalkSQSD middleware will forward each request and parameters to their
 appropriate jobs. The middleware will only process requests from the SQS daemon
 and will pass on others and so will not interfere with other routes in your
 application.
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -57,6 +57,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
+          bundler: latest
```
mullermp, line 60: What happens without this?

README.md
```diff
@@ -126,14 +126,35 @@ the `AWS_PROCESS_BEANSTALK_WORKER_REQUESTS` environment variable to `true` in
...
+may cause timeouts and incomplete executions.  
+
+To run jobs asynchronously, set the `AWS_PROCESS_BEANSTALK_WORKER_JOBS_ASYNC`
+environment variable to `true` in your worker environment.  Jobs will be queued
+in a ThreadPoolExecutor and the request will return a 200 OK immediately and the
+SQS message will be deleted and the job will be executed in the background.
+
+By default the executor will use the available processor count as the the
+max_threads.  You can configure the max threads for the executor by setting
+the `AWS_PROCESS_BEANSTALK_WORKER_THREADS` environment variable.
```
mullermp, line 150: Should we instead use/hijack `RAILS_MAX_THREADS` for the thread limit?

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -8,6 +8,8 @@ class ElasticBeanstalkSQSD
         def initialize(app)
           @app = app
           @logger = ::Rails.logger
+
+          init_executor if ENV['AWS_PROCESS_BEANSTALK_WORKER_JOBS_ASYNC']
```
mullermp, line 12: minor nit (up to you) but you could do something like:
```
return if ENV[...]
@executor = executor

def executor
  Concurrent::...
end
```
so that it's clear in initialization what instance variables may exist.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -28,25 +30,64 @@ def call(env)
...
+          threads = Integer(ENV.fetch('AWS_PROCESS_BEANSTALK_WORKER_THREADS',
+                                      Concurrent.available_processor_count || Concurrent.processor_count))
+          puts "Running with threads: #{threads}"
+          options = {
+            max_threads: threads,
+            max_queue: 1,
+            fallback_policy: :abort # Concurrent::RejectedExecutionError must be handled
+          }
+          @executor = Concurrent::ThreadPoolExecutor.new(options)
+          at_exit { shutdown }
```
mullermp, line 54: Is this the right place for this (on initialize)? But also, looking at the ThreadPoolExecutor docs, it looks like auto_terminate is true and will register an at_exit handler already. I'm assuming you only did this for logging?

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -28,25 +30,64 @@ def call(env)
...
+        end
+
+        # Execute a job using the thread pool executor
+        def _execute_job_parallel(request)
+          job_data = ::ActiveSupport::JSON.decode(request.body.string)
+          @executor.post(job_data) do |job|
+            @logger.debug("Executing job [in thread]: #{job['job_class']}")
+            ::ActiveJob::Base.execute(job)
+          end
+          [200, { 'Content-Type' => 'text/plain' }, ["Successfully queued job #{job_data['job_class']}"]]
```
mullermp, line 86: Should you also logger.info the messages in all of the success cases?

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -28,25 +30,64 @@ def call(env)
           periodic_task?(request) ? execute_periodic_task(request) : execute_job(request)
```
mullermp, line 30: Can we split up this into a method too?

mullermp (approved):

.github/workflows/ci.yml
alextwoods: This was just testing on the jruby issue.  I've removed now.

README.md
alextwoods: Talked offline - I think it conflates the number of threads used with Rails with the background threads and you may want to adjust them separately.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
alextwoods: yea, good idea.  Changed.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
alextwoods: this behavior seems to have changed in more recent versions of concurrent-ruby and without setting our own at_exit, we don't get clean shutdown.  I've kept this here because I'm not sure if theres a better place for it...

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
alextwoods: activejob already performs its own logging of these (both foreground and background).  Example:
```
[ActiveJob] [TestJob] [252646fa-067a-40f3-a1b4-00f40d604eb0] Performing TestJob (Job ID: 252646fa-067a-40f3-a1b4-00f40d604eb0) from Sqs(default)

```

sample-app/Gemfile
```diff
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 # Our gems
 # bundle config set local.aws-sdk-rails ../
-gem 'aws-sdk-rails', git: 'https://github.com/aws/aws-sdk-rails', branch: 'main'
+gem 'aws-sdk-rails', path: "../"
```
mullermp, line 5: Don't forget to change this back - you can set the branch and use bundler config local.

sample-app/app/jobs/test_job.rb
```diff
@@ -4,5 +4,8 @@ class TestJob < ApplicationJob
 
   def perform(*args)
     puts "Job performed with args: #{args}"
+    if args[0].is_a?(Hash) && args[0][:error]
```
mullermp, line 7: Probably not necessary, or if you want to keep, maybe a TestErrorJob or also place this code in async.

mullermp (approved):

PR #172, opened 2026-07-17
feat: Restrict which job classes the Elastic Beanstalk SQSD middleware will execute

sichanyoo:
_This PR summary / description was written with help of AI._

## Summary

The `ElasticBeanstalkSQSD` middleware deserializes the SQS Daemon's request body
and executes the class named in its `job_class` field, or the
`X-Aws-Sqsd-Taskname` header for periodic tasks. Previously any resolvable
constant would be resolved and run, so a request naming an arbitrary class could
cause that class to be executed. This change constrains what the middleware will
run.

## What this changes

`resolve_job_class` now applies three checks, ordered so the least trusting one
runs first:

1. **Name shape** - the name must match `JOB_CLASS_NAME_PATTERN`
   (`/\A[A-Z]\w*(::[A-Z]\w*)*\z/`), an optionally namespaced constant path and
   nothing else. Anything that cannot name a class is rejected before any lookup.
2. **Allowlist, compared as a String** - when `job_class_allowlist` is set, the
   raw name is matched against it *before* the constant is resolved.
3. **Resolve, then verify** - `Object.const_get(name, false)`, then a check that
   the result is a `Class` inheriting from `ActiveJob::Base`.

Both entry points share this method, so the periodic-task path is covered too.
That path is the more exposed of the two, since the name arrives in a request
header rather than a queue message.

Two details worth calling out for review:

- **Ordering.** Resolving a constant is not inert: under Zeitwerk it triggers
  autoloading, which runs the body of whichever file defines it. Checking the
  allowlist first means a disallowed name never reaches the resolver. To be
  precise about severity: this is defense-in-depth, not a code-execution fix.
  `const_get` does not evaluate arbitrary Ruby, and the previous code called only
  `is_a?`, `<`, and `.name` on the resolved class before raising. Checking a
  string against a list is simply cheaper than resolving a constant, and doing
  the cheap check first is the right order regardless.
- **`inherit: false`.** Each segment must be defined directly on the preceding
  one, so a name like `SomeJob::Foo` cannot resolve `Foo` from `SomeJob`'s
  ancestors. Without this, `ElasticBeanstalkJob::CallbackChain` resolves to
  `ActiveSupport::Callbacks::CallbackChain`.

The CodeQL suppression comment moved with the lookup into
`constantize_job_class` and now cites the pattern and allowlist that run before
it, in addition to the ancestry check after.

## Configuring the middleware

Configuration goes through a `.configure` block backed by a `Configuration`
class, which now lives in its own file at
`lib/aws/rails/middleware/elastic_beanstalk_sqsd/configuration.rb` (per review
feedback), mirroring the layout of the sibling `aws-activejob-sqs-ruby` change:

```ruby
# config/initializers/aws.rb
Aws::Rails::Middleware::ElasticBeanstalkSQSD.configure do |config|
  config.job_class_allowlist = [SomeJob, AnotherJob]
end
```

The class is reopened in the new file, so the constant remains
`ElasticBeanstalkSQSD::Configuration` and `.config` / `.configure` are unchanged.
The `require` sits in the `lib/aws-sdk-rails.rb` entry point alongside the
existing flat `require_relative` list, ahead of the middleware.

On keeping `Configuration` nested under the middleware rather than at the gem
module level: both patterns exist in the wild. `Rack::Attack::Configuration` is
nested under the middleware in its own file, which is what this does;
`SecureHeaders::Configuration` sits at the module level, but there the config is
the gem's entire surface and the middleware is an incidental delivery mechanism.
This gem has several unrelated features (logger wiring, encrypted credentials,
SDK instrumentation, notifications) to which `job_class_allowlist` is
meaningless, so scoping it to the middleware that owns it seemed right. Happy to
move it if you'd prefer the module-level shape.

Matching is by class name rather than object identity, so an allowlisted class is
still recognized after a Zeitwerk reload replaces the class object. The allowlist
is configured in code only, taking an array of `Class` objects or name `String`s.
This matches the existing convention in this gem, where behavioral settings are
set in an initializer and ENV vars are reserved for deployment toggles
(`AWS_PROCESS_BEANSTALK_WORKER_*`). The gem has no YAML config-file machinery, so
no file channel is added.

Default remains `nil`, allowing any `ActiveJob::Base` subclass, so existing
behavior is preserved.

## Errors

Invalid or disallowed job classes raise
`ElasticBeanstalkSQSD::InvalidJobClassError` instead of a broad `ArgumentError`,
so unrelated `ArgumentError`s are no longer swallowed by the middleware's rescue
clauses. Failures are logged and return a 500 response, as before.

`NameError` from an unknown constant is left to propagate to the existing
`rescue NameError, InvalidJobClassError` in the callers, preserving the current
500 behavior. Those rescues are scoped to resolution only, so a `NameError`
raised from inside a job's own `#perform` is not mislabeled as a
class-resolution failure.

## Tests

- Rejects classes that do not inherit from `ActiveJob::Base`.
- Rejects names that are not well-formed constant paths.
- Does not resolve the constant when the name is malformed, or when the
  allowlist excludes it (these pin the ordering).
- Does not resolve a disallowed periodic task name from the request header.
- Does not resolve a name through an ancestor of the namespace.
- Allows / rejects classes against a configured `job_class_allowlist`.
- Still accepts an allowlisted class after a simulated Zeitwerk reload replaces
  the class object.
- Does not mislabel a `NameError` raised from within a valid job.

Full suite: 102 examples, 0 failures. Rubocop clean.

The new ordering and `inherit: false` tests were checked against the previous
implementation and fail there, so they pin real behavior rather than passing
incidentally.

## Docs

- README: documents the `.configure` block, notes that the allowlist is checked
  before resolution and applies to periodic task names, and keeps the security
  note about restricting worker port 80 access on Docker-based EB platforms.
- CHANGELOG: entries lead with customer impact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Unreleased Changes
 ------------------
 
+* Issue - Validate that job classes inherit from `ActiveJob::Base` before execution in Elastic Beanstalk SQS middleware.
+* Feature - Add optional `job_class_allowlist` class attribute to `ElasticBeanstalkSQSD` middleware to restrict which job classes can be dispatched.
+
 5.1.0 (2024-12-05)
 ------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 Unreleased Changes
 ------------------
 
-* Issue - Only classes inheriting from `ActiveJob::Base` are executed by the `ElasticBeanstalkSQSD` middleware, preventing arbitrary classes named in an SQS message from being instantiated and run.
-* Feature - Add optional `job_class_allowlist` configuration to `ElasticBeanstalkSQSD` middleware to restrict which job classes can be dispatched.
+* Issue - Only classes inheriting from `ActiveJob::Base` are executed by the `ElasticBeanstalkSQSD` middleware, preventing arbitrary classes named in an SQS message from being instantiated and run. Job class and periodic task names are validated as constant paths and resolved without searching the namespace's ancestors, so a request cannot name a constant outside the intended namespace.
+* Feature - Adds a new configuration object for elastic beanstalk sqsd middleware, with an optional `job_class_allowlist` configuration to `ElasticBeanstalkSQSD` middleware to restrict which job classes can be dispatched. When set, the allowlist is checked before the class name is resolved, so an excluded class is never loaded.
 
 5.1.0 (2024-12-05)
 ------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 Unreleased Changes
 ------------------
 
-* Issue - Validate that job classes inherit from `ActiveJob::Base` before execution in Elastic Beanstalk SQS middleware.
-* Feature - Add optional `job_class_allowlist` class attribute to `ElasticBeanstalkSQSD` middleware to restrict which job classes can be dispatched.
+* Issue - Only classes inheriting from `ActiveJob::Base` are executed by the `ElasticBeanstalkSQSD` middleware, preventing arbitrary classes named in an SQS message from being instantiated and run.
+* Feature - Add optional `job_class_allowlist` configuration to `ElasticBeanstalkSQSD` middleware to restrict which job classes can be dispatched.
 
 5.1.0 (2024-12-05)
 ------------------
--- a/lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
+++ b/lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
@@ -177,7 +177,7 @@ module Aws
         end
 
         def validate_job_class!(name)
-          klass = name.constantize # lgtm[rb/code-injection]
+          klass = name.constantize # CodeQL [rb/code-injection] Mitigated by ActiveJob::Base ancestry check below
           unless klass.is_a?(Class) && klass < ::ActiveJob::Base
             raise ArgumentError, "#{name} is not a valid job class (must inherit from ActiveJob::Base)"
           end
--- a/lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
+++ b/lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
@@ -12,10 +12,12 @@ module Aws
       class ElasticBeanstalkSQSD # rubocop:disable Metrics/ClassLength
         # Configuration for the {ElasticBeanstalkSQSD} middleware.
         class Configuration
-          # @return [Array<Class>, nil] Optional list of job classes permitted
-          #   to be executed. When set, only classes in this list will be
-          #   dispatched. When nil, any class inheriting from ActiveJob::Base
-          #   is allowed.
+          # @return [Array<Class, String>, nil] Optional list of job classes
+          #   permitted to be executed. When set, only classes in this list
+          #   will be dispatched. When nil, any class inheriting from
+          #   ActiveJob::Base is allowed. Entries may be given as Class objects
+          #   or Strings; matching is by class name, so an allowlisted class is
+          #   still recognized after a Zeitwerk reload replaces the class object.
           attr_accessor :job_class_allowlist
         end
 
@@ -120,21 +122,24 @@ module Aws
           # Jobs queued from the SQS adapter contain the JSON message in the request body.
           job = ::ActiveSupport::JSON.decode(request.body.string)
           job_name = job['job_class']
-          validate_job_class!(job_name)
+          # Scope the rescue to resolution only. A NameError raised from inside
+          # the job's own #perform must not be mislabeled as a class-resolution
+          # failure, so ::ActiveJob::Base.execute runs outside the begin/rescue.
+          begin
+            resolve_job_class(job_name)
+          rescue NameError, InvalidJobClassError => e
+            return unresolved_job_class_response(job_name, e)
+          end
           @logger.debug("Executing job: #{job_name}")
           ::ActiveJob::Base.execute(job)
           [200, { 'Content-Type' => 'text/plain' }, ["Successfully ran job #{job_name}."]]
-        rescue NameError, InvalidJobClassError => e
-          @logger.error("Job #{job_name} could not resolve to a class that inherits from Active Job.")
-          @logger.error("Error: #{e}")
-          internal_error_response
         end
 
         # Execute a job using the thread pool executor
-        def _execute_job_background(request) # rubocop:disable Metrics/MethodLength
+        def _execute_job_background(request)
           job_data = ::ActiveSupport::JSON.decode(request.body.string)
           job_name = job_data['job_class']
-          validate_job_class!(job_name)
+          resolve_job_class(job_name)
           @logger.debug("Queuing background job: #{job_name}")
           @executor.post(job_data) do |job|
             ::ActiveJob::Base.execute(job)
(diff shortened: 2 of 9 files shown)


lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -156,6 +176,18 @@
           request.headers['X-Aws-Sqsd-Taskname'].present? && request.fullpath == '/'
         end
 
+        def validate_job_class!(name)
+          klass = name.constantize # CodeQL [rb/code-injection] Mitigated by ActiveJob::Base ancestry check below
```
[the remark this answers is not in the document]
sichanyoo: The user-provided value gets validated afterwards; this warning can be safely ignored.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -4,7 +4,17 @@ module Aws
...
-      class ElasticBeanstalkSQSD
+      #
+      # @example Restrict job dispatch to specific classes
+      #   Aws::Rails::Middleware::ElasticBeanstalkSQSD.job_class_allowlist = [SendReceiptJob, ProcessOrderJob]
+      class ElasticBeanstalkSQSD # rubocop:disable Metrics/ClassLength
+        # Optional list of job classes permitted to be executed. When set, only
+        # classes in this list will be dispatched. When nil, any class inheriting
+        # from ActiveJob::Base is allowed.
+        class << self
+          attr_accessor :job_class_allowlist
```
jterapin: Hmm. I'm not completely sold on this class variable. I think its worth checking what other Rail Middleware sets their configuration and follow that pattern.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -4,7 +4,17 @@ module Aws
   module Rails
     module Middleware
       # Middleware to handle requests from the SQS Daemon present on Elastic Beanstalk worker environments.
-      class ElasticBeanstalkSQSD
+      #
```
jterapin, line 7: nit: remove

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -22,7 +32,9 @@ def call(env)
 
           @logger.debug('aws-sdk-rails middleware detected call from Elastic Beanstalk SQS Daemon.')
 
-          # Only accept requests from this user agent if it is from localhost or a docker host in case of forgery.
+          # Best-effort source check. On Docker-based EB platforms this is unreliable
+          # because all traffic arrives via the docker bridge. Use job_class_allowlist
+          # and network-level controls (security groups) as the primary safeguards.
           unless request.local? || sent_from_docker_host?(request)
```
jterapin, line 63: Does this info surface in the generated docs? Might worth checking that

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -78,39 +90,47 @@ def _execute_job_now(request)
           # Jobs queued from the SQS adapter contain the JSON message in the request body.
           job = ::ActiveSupport::JSON.decode(request.body.string)
           job_name = job['job_class']
+          validate_job_class!(job_name)
           @logger.debug("Executing job: #{job_name}")
           ::ActiveJob::Base.execute(job)
           [200, { 'Content-Type' => 'text/plain' }, ["Successfully ran job #{job_name}."]]
-        rescue NameError => e
+        rescue NameError, ArgumentError => e
```
jterapin: I'm guessing this is handling a case where a JobClass is not allowed due to the allowlist? Is it better to raise a custom error than a broad `ArgumentError`? 🤔

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 Unreleased Changes
 ------------------
 
+* Issue - Validate that job classes inherit from `ActiveJob::Base` before execution in Elastic Beanstalk SQS middleware.
```
jterapin: I would use a similar changelog entry as I suggested in the other pr.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
jterapin: Example: https://github.com/github/secure_headers/blob/main/lib/secure_headers/configuration.rb

jterapin (changes requested):
Nice! Looks like its heading to a good direction. I will apply similar feedbacks as I had in other PR: updating pr description.

CHANGELOG.md
sichanyoo: Changed as suggested.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
sichanyoo: Added custom error as suggested.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
sichanyoo: It doesn't surface in generated docs afaik but README covers it.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
sichanyoo: Added more doc comment underneath it which happens to resolve nit

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
sichanyoo: Replaced bare attr_accessor with a lightweight Configuration class + .configure { |c| } / .config API. Follows the yield(config) Rails idiom

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -156,6 +225,25 @@ def periodic_task?(request)
           request.headers['X-Aws-Sqsd-Taskname'].present? && request.fullpath == '/'
         end
 
+        # Resolves +name+ to a job class, raising InvalidJobClassError if it
+        # does not name a class inheriting from ActiveJob::Base or is not in the
+        # configured allowlist. Returns the resolved class so callers can reuse
+        # it without resolving the name a second time.
+        def resolve_job_class(name)
+          klass = name.constantize # CodeQL [rb/code-injection] Mitigated by ActiveJob::Base ancestry check below
```
richardwang1124: Same comment in the other PR, should we consider checking the name against the allowlist first?

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
jterapin: Could this be extracted to its own file? to follow similar pattern as the other pr?

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 Unreleased Changes
 ------------------
 
+* Issue - Only classes inheriting from `ActiveJob::Base` are executed by the `ElasticBeanstalkSQSD` middleware, preventing arbitrary classes named in an SQS message from being instantiated and run.
+* Feature - Add optional `job_class_allowlist` configuration to `ElasticBeanstalkSQSD` middleware to restrict which job classes can be dispatched.
```
jterapin: Suggestion: Since we are introducing a configuration object, i think we can frame it a little differently here. This sounds like an addition to an existing one when it isn't.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
sichanyoo: Added similar change as other PR.

CHANGELOG.md
sichanyoo: Changed wording as suggested.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
sichanyoo: Extracted config to its own file.

lib/aws/rails/middleware/elastic_beanstalk_sqsd.rb
```diff
@@ -156,6 +218,48 @@
...
+        end
+
+        # Resolves an already-validated name. +inherit+ is false so that each
+        # segment must be defined directly on the preceding one: a name such as
+        # 'SomeJob::Foo' cannot resolve Foo from SomeJob's ancestors when
+        # SomeJob itself does not define it.
+        def constantize_job_class(name)
+          # CodeQL [rb/code-injection] Name is validated against JOB_CLASS_NAME_PATTERN
+          # and the allowlist above, and the result is checked for ActiveJob::Base ancestry below.
+          Object.const_get(name, false)
```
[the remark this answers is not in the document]
sichanyoo: Ignore; we allow empty or nil allowlist for users that don't care to use it. Requiring it would be a breaking change.

jterapin (approved):
lgtm

